PR #6, opened 2017-10-23
Simplify for "boolean zen"

tannewt:
Its simpler to assign a boolean directly rather than doing it through an if statement like:
```
if foo:
    bar = True
else:
    bar = False
```

Instead, just do `bar = foo`


ladyada:
i know its totally weird, but i'd like this beginner guide to use if/else because we expect people to change the behaviour based on the switch...OK to reject?

tannewt:
Yup! I'll follow up with an indentation fix.

PR #8, opened 2017-10-24
added Godron Cole MP3 Player

caternuson:
@ladyada Arduino code for Gordon Cole MP3 Player project


ladyada:
ooh rather than a submodule, please commit directly?

PR #19, opened 2017-11-22
Examples from Intro Guide using cpx express API

kattni:
These are examples of the same code in the Intro to Circuit Playground Express learn guide Circuit Python section using the `cpx` API.


ladyada:
looks awesome. reviewed and approved :)

PR #24, opened 2017-12-12
Use lower case variables with underscores.

tannewt:


tannewt:
Nope, I'm at the library without one. I was hoping you or @kattni would be kind enough to test.

jedgarpark:
Tested, works just fine.

PR #121, opened 2018-04-10
Include both RGB and RGBW examples in earrings code

klajo:
Hi,

I've extended the Gemma_Hoop_Earrings.py:

The Gemma_Hoop_Earrings.py example is prepared both for RGB and
RGBW (with an extra white) NeoPixel rings.  The RGBW requires the bpp=4
option (default is bpp=3) and the number of elements in the color value
list is different: [R, G, B] vs [R, G, B, W].

In the interest of keeping the code as simple as possible I just added a version for RGBW rings that's "commented out" in the code, instead of trying to make something that adapts depending on the setting of some variable.

Hope this is useful for those who've bought a RGBW ring, instead of RGB.

Cheers,
Klas

### DIFF
--- a/Gemma_Hoop_Earrings/Gemma_Hoop_Earrings.py
+++ b/Gemma_Hoop_Earrings/Gemma_Hoop_Earrings.py
@@ -9,23 +9,35 @@ try:
 except ImportError:
 	import random
 
+# The example is prepared both for RGB and RGBW (with an extra white)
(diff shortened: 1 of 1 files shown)


kattni:
Thank you for the PR! However, we've documented in the guide that the project is designed for RGB NeoPixel rings. We will not be adding RGBW support to the code.

klajo:
Aha, ok.

PR #122, opened 2018-04-10
Change from write to show in earrings code

klajo:
Hi,

Another change to the Gemma earrings example:

The NeoPixel API docs now state that show() should be used instead of
write().  Change the Gemma_Hoop_Earrings.py example.

    http://circuitpython.readthedocs.io/projects/neopixel/en/latest/api.html

Cheers,
Klas


klajo:
No problem!

PR #123, opened 2018-04-10
Hoop earrings no auto write

klajo:
Hi,

This expands on the change in pull request #121 and turns off auto_write in the earrings example to speed up the program.

Cheers,
Klas


kattni:
We are not adding RGBW support to this project - the project was designed for RGB NeoPixel rings. Due to merge conflicts, I have added this manually in #607. Thanks for the suggestion!

PR #128, opened 2018-04-13
Update stats.py

dirtydavemcgee:
Improves readability and adds cpu, disk and memory stats back.

### DIFF
--- a/Pi_Hole_Ad_Blocker/stats.py
+++ b/Pi_Hole_Ad_Blocker/stats.py
@@ -96,18 +96,24 @@ while True:
     except:
       time.sleep(1)
       continue
+        
+#DirtyDaveMcGee edit to improve readability
+#192.168.x.x addresses were being cut off on my display
+#and i wanted to keep the cpu, disk, and memory stats
 
-    draw.text((x, top),       "IP: " + str(IP) + "( " + HOST + ")",  font=font, fill=255)
+    draw.text((x, top),	      str(HOST), font=font, fill=255)
     draw.text((x, top+8),     "Ads Blocked: " + str(ADSBLOCKED), font=font, fill=255)
     draw.text((x, top+16),    "Clients:     " + str(CLIENTS),  font=font, fill=255)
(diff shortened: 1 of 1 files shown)


dirtydavemcgee:
Comment for clarity
The main reason for this change was the overlapping of the IP address when using anything not on a 10.0.0.x network. My home network is 192.168.10.xxx and would not display correctly.

Pi_Hole_Ad_Blocker/stats.py
```diff
@@ -96,18 +96,24 @@
     except:
       time.sleep(1)
       continue
+        
+#DirtyDaveMcGee edit to improve readability
+#192.168.x.x addresses were being cut off on my display
+#and i wanted to keep the cpu, disk, and memory stats
```
dhalbert: Don't bother to annotate your changes. If they work for all cases, just make the changes.

Pi_Hole_Ad_Blocker/stats.py
```diff
@@ -96,18 +96,24 @@
...
 
-    draw.text((x, top),       "IP: " + str(IP) + "( " + HOST + ")",  font=font, fill=255)
+    draw.text((x, top),	      str(HOST), font=font, fill=255)
     draw.text((x, top+8),     "Ads Blocked: " + str(ADSBLOCKED), font=font, fill=255)
     draw.text((x, top+16),    "Clients:     " + str(CLIENTS),  font=font, fill=255)
     draw.text((x, top+24),    "DNS Queries: " + str(DNSQUERIES),  font=font, fill=255)
```
dhalbert: Could you change the above to use `.format()` or '%' instead of string concatenation?

dirtydavemcgee: If I had the slightest clue as to what I was doing, I'd guess it was possible. This is a copy paste hack based on my non-understanding of python. It works for me. I believe the code was in a tutorial on adafruit.com

ladyada:
hiya we're going to keep the default stats.py as it is - but the code is there for you to customize as you need so keep hacking :D

PR #134, opened 2018-04-25
fix possible overrun & add circuitpy3 compatibility

ladyada:
tested on a metro m4 w/cpy 3a6


dhalbert:
On the error cases
```
    if frame_len != 28:
        continue
```
and
```
    if check != checksum:
        continue
```
I thought you would need to clear out the buffer (set it to `[]` before `continue`) because in both of these cases it seems to me you have read bad data and need to discard what's read so far, rather than leave it in the buffer. But I didn't study the datasheet in detail.

ladyada:
added!

PR #178, opened 2018-05-23
Code for CP Made Easy on CPX

kattni:


kattni:
Linted locally successfully.

PR #208, opened 2018-05-31
Added import NeoPixel - requires guide update

kattni:


kattni:
Guide updated.

kattni:
I was thinking pylint would complain about the import in the middle of the code, but that's still the top, so it's fine. Good call. I moved it.

PR #230, opened 2018-06-25
Update 3D_Printed_Guardian_Sword.py

Chompworks:
got rid of some magic numbers from the code, since range() in python ensures no overlapping


ladyada:
@djecken hey wanna take a look? this change seems good to me :)

PR #232, opened 2018-06-26
don't import specific pins

dhalbert:
Arcade button example was importing specific pins, including `board.D13`, which wasn't used. A user in Discord said mu complained about this, and it was kind of confusing. Changed to `import board` and just use pins.

Now if example is edited, import list does not need to be changed.

### DIFF
--- a/Arcade_Button_Control_Box/Arcade_Button_Control_Box.py
+++ b/Arcade_Button_Control_Box/Arcade_Button_Control_Box.py
@@ -3,13 +3,13 @@ import time
 import digitalio
 from adafruit_hid.keyboard import Keyboard
 from adafruit_hid.keycode import Keycode
-from board import D13, D12, D11, D10, D9, D6, D5, A0, A1, A2, A3, A4, A5
+import board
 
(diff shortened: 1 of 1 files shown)


kattni:
@dhalbert Travis failed, one of the variables is undefined.

Arcade_Button_Control_Box/Arcade_Button_Control_Box.py
```diff
@@ -3,13 +3,13 @@
...
 from adafruit_hid.keycode import Keycode
-from board import D13, D12, D11, D10, D9, D6, D5, A0, A1, A2, A3, A4, A5
+import board
 
 # A simple neat keyboard demo in circuitpython
 
 # The button pins we'll use, each will have an internal pullup
-buttonpins = [D12, D11, D10, D9, D6, D5]
-ledpins = [A0, A1, A2, A3, A4, A5]
+buttonpins = [board.D12, board.D11, board.D10, board.D9, board.D6, board.D5]
```
kattni: You did not include D13 here.

PR #245, opened 2018-07-10
Audio out essentials

kattni:
Audio code plus some of the M4 updates. (Unintentionally combined, but it's needed anyway so I'm leaving it together.)

### DIFF
--- a/CircuitPython_Essentials/CircuitPython_Audio_Out_Wave.py
+++ b/CircuitPython_Essentials/CircuitPython_Audio_Out_Wave.py
@@ -1,0 +1,26 @@
+import time
+import audioio
+import board
+import digitalio
+
+button = digitalio.DigitalInOut(board.A1)
+button.direction = digitalio.Direction.INPUT
+button.pull = digitalio.Pull.UP
+
+wave_file = open("StreetChicken.wav", "rb")
+wave = audioio.WaveFile(wave_file)
+audio = audioio.AudioOut(board.A0)
+
+while True:
+    audio.play(wave)
+    t = time.monotonic()
+    while time.monotonic() - t < 6:
+        pass
+    audio.pause()
+    print("Waiting for button press to continue!")
+    while button.value:
+        pass
+    audio.resume()
+    while audio.playing:
+        pass
+    print("Done!")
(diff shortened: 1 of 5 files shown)


CircuitPython_Essentials/CircuitPython_Audio_Out_Wave.py
```diff
@@ -0,0 +1,26 @@
...
+
+wave_file = open("StreetChicken.wav", "rb")
+wave = audioio.WaveFile(wave_file)
+audio = audioio.AudioOut(board.A0)
+
+while True:
+    audio.play(wave)
+    t = time.monotonic()
+    while time.monotonic() - t < 6:
+        pass
```
dhalbert: Does `time.sleep(6)` work here? Would be simpler.

kattni: Yes apparently it does. I think it was done that way to keep the code non-blocking - I copied it from a Crickit example which I assume the idea was to be able to do all of the other Crickit things while the audio plays. 

Do you think it's worth doing it that way for this example so it's also non-blocking and therefore adaptable for other things with more going on? Or go with simple since it's for the essentials guide?

ladyada: i prefer the above w/monotonic cause it's more flexible, and its not *much* more complex!

kattni: Sounds good.

dhalbert: I don't think `time.sleep()` will block the playing, but it's true that using the example, you can do other things inside the loop, like poll buttons, etc. You could put a comment above the `pass`  in to the effect of "You could check buttons here or do other things".

dhalbert: ^ is that what you mean @ladyada ?

kattni: Right. Blocking v non-blocking. This is currently more adaptable to doing other things while the file plays. With `time.sleep()` it wouldn't be.

dhalbert: k - great: could you just add a comment `# Do something else here while the audio plays.`

kattni: Yep! Separated the block, and added the comment. It'll be explained in the guide page as well.

PR #247, opened 2018-07-12
Add code for CP basic data structures guide

dastels:


CircuitPython_101/basic_data_structures/song_book/debouncer.py
```diff
@@ -0,0 +1,105 @@
...
+class Debouncer(object):
+    """Debounce an input pin"""
+
+    DEBOUNCED_STATE = 0x01
+    UNSTABLE_STATE = 0x02
+    CHANGED_STATE = 0x04
+
+
+    def __init__(self, pin, mode=None, interval=0.010):
+        """Make am instance.
```
kattni: I believe you mean `Make an instance.`

CircuitPython_101/basic_data_structures/play_note/code.py
```diff
@@ -0,0 +1,31 @@
...
+G4     = 392
+G_SH_4 = 415
+A4     = 440
+A_SH_4 = 466
+B4     = 493
+
+def play_note(note):
+    if note[0] != 0:
+        pwm = pulseio.PWMOut(board.D12, duty_cycle = 0, frequency=note[0])
+        pwm.duty_cycle = 0x7FFF
```
kattni: I'm not sure how basic you intend for this to be, but using hex here is not as beginner friendly. If this isn't meant to be a beginner guide, ignore this suggestion.

CircuitPython_101/basic_data_structures/song_book/code.py
```diff
@@ -0,0 +1,87 @@
...
+                              (G4, 0.5), (G4, 0.5), (G4, 0.5), (G4, 0.5), (G4, 0.5), (G4, 0.5), (0, 0.5),
+                              (G4, 0.5), (G4, 0.5), (G4, 0.5), (D4, 0.5), (E4, 0.5), (E4, 0.5), (D4, 0.5), (0, 0.5),
+                              (B4, 0.5), (B4, 0.5), (A4, 0.5), (A4, 0.5), (G4, 0.5), (0, 0.5)]
+           }
+# pylint: enable=line-too-long
+
+def play_note(note):
+    if note[0] != 0:
+        pwm = pulseio.PWMOut(board.D12, duty_cycle = 0, frequency=note[0])
+        pwm.duty_cycle = 0x7FFF
```
kattni: Same as above.

CircuitPython_101/basic_data_structures/song_list/code.py
```diff
@@ -0,0 +1,33 @@
...
+A_SH_4 = 466
+B4     = 493
+
+twinkle = [(C4, 0.5), (C4, 0.5), (G4, 0.5), (G4, 0.5), (A4, 0.5), (A4, 0.5), (G4, 0.5), (0, 0.5),
+           (F4, 0.5), (F4, 0.5), (E4, 0.5), (E4, 0.5), (D4, 0.5), (D4, 0.5), (C4, 0.5)]
+
+def play_note(note):
+    if note[0] != 0:
+        pwm = pulseio.PWMOut(board.D12, duty_cycle = 0, frequency=note[0])
+        pwm.duty_cycle = 0x7FFF
```
kattni: Same as above.

CircuitPython_101/basic_data_structures/play_note/code.py
dastels: I marked it intermediate and I'd consider it to be post-beginner. This is the first of a series I have in mind. The goal is to help folks take the next step. Now with the M4 boards it's possible to do go beyond simple scripts. This series aims to help people do that.

Specific to this case, having to specify duty cycle as an unsigned 16 bit int isn't very beginner friendly (whether in hex or decimal). It would be better done as an int between 0 & 100 (inclusive) since duty cycle is generally thought of as a percentage, and not as the comparison value for the 16 bit PWM counter. Actually supporting both would be ideal.

dhalbert: I added `.fraction` to `adafruit_seesaw.PWMOut`, and we can do the same for `pulseio.PWMOut` here. It was late in the game so I didn't add it here. `.fraction` is between 0.0 and 1.0. I'll open an issue.

PR #249, opened 2018-07-14
Code for Crickit+AdafruitIO guide

dastels:


jerryneedell:
Forgot to add that - I did finally get this work on the M0 ATWINC   -- the problem was an out of date version of the adafruit-io-arduino library.. All good now

TheKitty:
Awesome, thanks Jerry. There appears to be something wrong with Travis lumping the whole of the projects in the learning system into one report. This doesn't allow authors of individual learn projects to correct their code and get a clean report if there are issues elsewhere. Is anyone else seeing this? I think @dastels was seeing this too.

ladyada:
mike, they need to do another git pull to get the latest code - see if that helps

dastels:
Yes. I updated my projects that were failing the new pylint rules.  The special flavor of frustration in this case was that my PR was C based.  No python. And it got blocked because of python projects. :(  Imperfect, but understandable.

PR #252, opened 2018-07-17
Techno tiki

mikeysklar:
Ported all three Arduino Learn Examples to CircuitPython.


mikeysklar:
ladyada - you bet. I have some things to correct.

ladyada:
i think if you run within the learn_guide top level folder pylint will see the .pylintrc and then you can run `pylint my_folder/script.py`

mikeysklar:
@ladyada while travis still fails it is due to other projects in the Learn archive. Techno_Tiki builds clean no messages.

ladyada:
PR is fine - not sure why CI/push is complaining!

mikeysklar:
@ladyada - I can assembled the 3D Printed Flag Circuit again and correct the tuple warning we are getting from travis-ci/push. At least that will be one less bit of noise. 

112.40s$ ./pylint_check
************* Module 3D_Printed_LED_Microphone_Flag
3D_Printed_LED_Microphone_Flag/3D_Printed_LED_Microphone_Flag.py 137: Consider using tuple unpacking for swapping variables (consider-swap-variables)

ladyada:
no need, we downgraded pylint so it no longe complains!

tannewt:
Globals are ok in my opinion because its a script and not a library. The pythonic way would be to wrap stuff in an object (aka `class`).

Issue #254, opened 2018-07-18
set pylint to 1.92 and undo changes to pylintrc


ladyada:
hey dan, pylint updated! please tweak travis so it uses 1.9.2 only and revert this commit
https://github.com/adafruit/Adafruit_Learning_System_Guides/commit/6fc1cd11a3ac795329b3dd7a583d495ba9942012
(i disabled consider-swap-variables)

dhalbert:
Fixed by #255.

PR #255, opened 2018-07-18
Pylint force 1.9.2

dhalbert:
`pip install --force-reinstall pylint=1.9.2` makes sure the version is 1.9.2 whether or not it's already installed.


dhalbert:
Fixes #254.

PR #261, opened 2018-07-19
CP code for I2S, pin script

kattni:


kattni:
Script has been tested by multiple people. I tested all the pin combinations that Feather M0 returned and all of them work successfully.

PR #282, opened 2018-07-31
Create cpx-cap-touch.py

TheKitty:
first run for cap touch

### DIFF
--- a/Make_It_Sense/cpx-cap-touch.py
+++ b/Make_It_Sense/cpx-cap-touch.py
@@ -1,0 +1,20 @@
+from adafruit_circuitplayground.express import cpx
+
+# NeoPixel blank out
+cpx.pixels.fill( (0, 0, 0) )
+
+# Set sensitivity of all touch pads
+cpx.adjust_touch_threshold(200)
+
+# Check for touch, if so light the nearby NeoPixel green
+while True:
+    if cpx.touch_A4:
+        cpx.pixels[0] = (0, 30, 0)
+    if cpx.touch_A5:
+        cpx.pixels[1] = (0, 30, 0)
+    if cpx.touch_A6:
+        cpx.pixels[3] = (0, 30, 0)
+    if cpx.touch_A7:
+        cpx.pixels[4] = (0, 30, 0)
+    if cpx.button_a:  # blank lights on Button A
+        cpx.pixels.fill( (0, 0, 0) )


Make_It_Sense/cpx-cap-touch.py
```diff
@@ -0,0 +1,20 @@
...
+    if cpx.touch_A4:
+        cpx.pixels[0] = (0, 30, 0)
+    if cpx.touch_A5:
+        cpx.pixels[1] = (0, 30, 0)
+    if cpx.touch_A6:
+        cpx.pixels[3] = (0, 30, 0)
+    if cpx.touch_A7:
+        cpx.pixels[4] = (0, 30, 0)
+    if cpx.button_a:  # blank lights on Button A
+        cpx.pixels.fill( (0, 0, 0) )
```
kattni: Same as above.

Make_It_Sense/cpx-cap-touch.py
```diff
@@ -0,0 +1,20 @@
+from adafruit_circuitplayground.express import cpx
+
+# NeoPixel blank out
+cpx.pixels.fill( (0, 0, 0) )
+
+# Set sensitivity of all touch pads
+cpx.adjust_touch_threshold(200)
```
kattni: Were you running into threshold issues or simply wanted to include the threshold change in the event that people need to alter it later?

Make_It_Sense/cpx-cap-touch.py
```diff
@@ -0,0 +1,20 @@
+from adafruit_circuitplayground.express import cpx
+
+# NeoPixel blank out
+cpx.pixels.fill( (0, 0, 0) )
```
kattni: You don't need the extra spaces around the `(0, 0, 0)`. It can be `((0, 0, 0))`.

Make_It_Sense/cpx-cap-touch.py
TheKitty: Just to show it can be done

Make_It_Sense/cpx-cap-touch.py
TheKitty: just to emphasize its a tuple

Make_It_Sense/cpx-cap-touch.py
kattni: Ok!

Make_It_Sense/cpx-cap-touch.py
kattni: Pylint isn't complaining, so we can leave it as-is, but it's not how we've done it previously. If you'd like to leave it, it's up to you. If you'd rather it be consistent, then consider changing it.

PR #299, opened 2018-08-17
first commit Adabot Operation Game code

jedgarpark:


Adabot_Operation_Game/adabot_operation_game.py
```diff
@@ -0,0 +1,42 @@
...
+CYAN = (0, 10, 10)
+GREEN = (0, 20, 0)
+YELLOW = (10, 10, 0)
+ORANGE = (15, 5, 0)
+RED = (20, 0, 0)
+WHITE = (3, 3, 3)
+
+COLORS = [MAGENTA, VIOLET, BLUE, CYAN, GREEN, YELLOW, ORANGE, RED, WHITE]
+
+cpx.pixels.fill((WHITE))
```
kattni: Since the variables you created have one set of parens already, I'm not sure that two sets are needed here in this line. Try `cpx.pixels.fill(WHITE)` to see if it works.

Adabot_Operation_Game/adabot_operation_game.py
```diff
@@ -0,0 +1,42 @@
...
+
+cap_pins = (board.A1, board.A2, board.A3, board.A4, board.A5,
+            board.A6, board.A7)
+touch_pads = []
+for i in range(7):
+    touch_pads.append(touchio.TouchIn(cap_pins[i]))
+for touch_pad in touch_pads:
+    touch_pad.threshold = 3500  # adjust value to fine-tune touch threshold
+
+MAGENTA = (10, 0, 10)
```
kattni: Is there a reason you chose to go with lower numbers in the tuples instead of full values and a line to lower the brightness overall? This is a perfectly excellent way to do it, however, I wonder if making the tuple values higher and allowing for brightness control might be good for people who build this project in case they want the lights to be brighter. Easier to update a brightness value on the fly than a series of tuples. But this is entirely a matter of opinion, and does not need to be changed!

Adabot_Operation_Game/adabot_operation_game.py
```diff
@@ -0,0 +1,42 @@
...
+while True:
+    for i in range(7):
+        # uncomment block to check the raw touch pad values
+        # print("raw %s value for pad " % i)
+        # print(touch_pads[i].raw_value)
+        # time.sleep(.5)
+
+        if touch_pads[i].value:
+            # print("Touched %s" % i)  # uncomment for debugging
+            cpx.pixels.fill((RED))
```
kattni: Same as above, try without the extra parens.

Adabot_Operation_Game/adabot_operation_game.py
```diff
@@ -0,0 +1,42 @@
...
+            cpx.play_tone(660, 0.7)
+            cpx.pixels.fill((COLORS[i]))
```
kattni: This may or may not need the extra parens as well. Try without and see!

Adabot_Operation_Game/adabot_operation_game.py
jedgarpark: great catch, thanks. I dropped the extra parens and it works. Will re-commit w that change.

Adabot_Operation_Game/adabot_operation_game.py
jedgarpark: ditto, works w/o extras

Adabot_Operation_Game/adabot_operation_game.py
jedgarpark: roger that.

Adabot_Operation_Game/adabot_operation_game.py
jedgarpark: great point! I'm going to leave this one as is in this case, so people can look at the color blending/brightness equation in a different way.

kattni: Sounds good!

kattni:
Thanks for updating!

PR #315, opened 2018-09-05
Trinket question block sound jewelry

mikeysklar:
Ported to CircuitPython. Created new Fritzing Diagram due to required pin changes for PWM access.


mikeysklar:
@ladyada - Trinket "Question Mark" Sound Jewelry is ready for your review. There was a failed check on another chunk of code in the archive "ufo/ufo.py" which was importing board and claims to be unused (but is used on line: 14 as board. NeoPixel).

ladyada:
bizarre, it passed before ??

ladyada:
ok i fixed in a patch, will re-bump

PR #330, opened 2018-09-14
first commit Milk Jug Glow Skull

jedgarpark:


Milk_Jug_Glow_Skull/milk_jug_glow_skull.py
```diff
@@ -0,0 +1,63 @@
...
+        led.value = False
+        time.sleep(speed)
+        led.value = True
+        time.sleep(speed)
+
+def play_waves(file_num):
+    wave_file = open(wave_files[file_num], "rb")
+    wave = audioio.WaveFile(wave_file)
+    audio.play(wave)
+
```
ladyada: you never close the file or wait for it to finish?

Milk_Jug_Glow_Skull/milk_jug_glow_skull.py
```diff
@@ -0,0 +1,63 @@
...
+wave_files = ["alex_deepgrowl1.wav", "alex-highgrowl1.wav", "alex-squeal1.wav",
+              "toni-deepgrowl.wav","toni-highgrowl2.wav","toni-pigsqueal.wav",
+              "toni-pitchedscream2.wav"]
+audio = audioio.AudioOut(board.A0)
+
+while True:
+    for k in range(7):
```
ladyada: als so this should be range(len(wave_files)) in case they add/remove

Milk_Jug_Glow_Skull/milk_jug_glow_skull.py
jedgarpark: I found you can do some fun things by setting the times very low and it just hits the start of each scream. Is this ok to let it start new ones, interrupting the old ones?

jedgarpark:
I removed the underscore in front of variable 'j' since it made pylint happy but didn't actually run on the board.

Milk_Jug_Glow_Skull/milk_jug_glow_skull.py
ladyada: :/ its kinda bad code practice, if you wanted to do this have it play the audio, then time.delay() and stop playing the audio before returning

ladyada:
ok- are you asking me to debug it?

jedgarpark:
I was seeing a pylint comment `Milk_Jug_Glow_Skull/milk_jug_glow_skull.py 43: Unused variable 'j' (unused-variable)` in the line `for j in range(times):` I found you can make pylint happy by adding the underscore in front of j, but it won't actually run on the board. I don't think this is a serious pylint error, right? If it is, I'll need to learn a different way to iterate through a loop a number of times.

ladyada:
ooh replace it with just '_'!

jedgarpark:
Wizard!

Milk_Jug_Glow_Skull/milk_jug_glow_skull.py
jedgarpark: Got it. I've added the while `audio.playing: pass` line to fix

Milk_Jug_Glow_Skull/milk_jug_glow_skull.py
jedgarpark: changed it, thanks.

ladyada:
ok dont forget to close the file when done!

jedgarpark:
I can't find anything on closing the wave file after playback, could you please remind me?

ladyada:
you can use 'with:' to auto-close the file
image
https://learn.adafruit.com/crickit-powered-owl-robot/code-with-circuitpython

or just call file.close() after its done playing

ladyada:
ok done!

jedgarpark:
Thanks for the help on the code for file closing!
Not embarrassed at all that it was in a previous project of mine :)

ladyada:
its all practice! you'll never forget again :D

PR #341, opened 2018-09-27
first commit of HalloWing Jump Scare Trap

jedgarpark:


HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
```diff
@@ -0,0 +1,152 @@
...
+# Setup servo
+servo = pulseio.PWMOut(board.D4, frequency=50)
+servo_min = 0.6
+servo_mid = 1.5
+servo_max = 2.45
+servo_time = 2.0
+# Setup cap touch button
+ready_button = touchio.TouchIn(board.TOUCH1)
+# Create a function to simplify setting PWM duty cycle for the servo:
+def servo_duty_cycle(pulse_ms, frequency=50):
```
ladyada: can you use the adafruit_motor library instead? it does all this servo stuff for ya!

jedgarpark: I think I had it for motor library at first and then switched it for some reason I'm now forgetting. I'll switch it back to that and make sure everything is still working.

ladyada: thanks  :)

jedgarpark: I must have imagined all that. It works great w the motor library code, I'll clean it up and re-commit.

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
```diff
@@ -0,0 +1,148 @@
...
+PIR_PIN = board.SENSE   # Pin port connected to PIR sensor output wire.
+# Setup digital input for PIR sensor:
+pir = digitalio.DigitalInOut(PIR_PIN)
+pir.direction = digitalio.Direction.INPUT
+# Setup digital output for LED:
+led = digitalio.DigitalInOut(LED_PIN)
+led.direction = digitalio.Direction.OUTPUT
+# Setup servo
+# servo = pulseio.PWMOut(board.D4, frequency=50)
+pwm = pulseio.PWMOut(board.D4, duty_cycle=2 ** 15, frequency=50)
```
ladyada: i dont think you need to do the dutycycle/frequency part, try removing those?

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
```diff
@@ -0,0 +1,148 @@
...
+    servo.angle = 90
+
+# Set servo to ready position
+servo_ready()
+# Function for playing wav file, releasing servo
+def play_wave():
+    wave_file = open("hiss01.wav", "rb")  # open a wav file
+    wave = audioio.WaveFile(wave_file)
+    audio.play(wave)  # play the wave file
+    while audio.playing:  # turn on LED, turn servo
```
ladyada: i think you want to move these three lines above the while loop, and then just 'pass' within the loop, not continuously print out Motion Detected right?

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
```diff
@@ -0,0 +1,148 @@
...
+        led.value = True
+        servo_release()
+        print('Motion detected!')
+    wave_file.close()  # close the wav file
+# Setup audio out pin
+audio = audioio.AudioOut(board.A0)
+
+# tone player setup for status beeps
```
ladyada: replace this with simpleio's tone()?
https://learn.adafruit.com/using-piezo-buzzers-with-circuitpython-arduino/circuitpython#making-tones-with-simpleio-3-19

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
jedgarpark: I'm not sure how to define the servo object on the next line without the pwm line.

jedgarpark: Oh, got it, this works as you suggested now.

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
jedgarpark: I tried this, but seems like the simplio.tone won't work on A0 because it's looking for PWM. I got this error: `TypeError: extra positional arguments given` Do you have any suggestions? Otherwise I'll leave it as it's currently being done.

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
jedgarpark: Perfect, this worked great.

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
ladyada: @kattni can you help @jedgarpark with this? just should just work i thought we added it to simpleio

jedgarpark: @kattni helped me with this earlier (thanks!) and found that currently the tone code doesn't want to talk to the A0 pin.

ladyada: it should
https://github.com/adafruit/Adafruit_CircuitPython_SimpleIO/blob/master/simpleio.py#L52

kattni: Looking into this.

kattni: This is because `tone` hasn't been updated to support the changes to `audiout` introduced in 3.x.

ladyada: oooh @kattni plz put that on your todo list for this week to fix! this code can go live for now until then, then come back and make sure this works

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
ladyada: did you actually commit this change?

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
kattni: Ok! Added to the list. I'll keep @jedgarpark posted.

HalloWing_Jump_Scare_Trap/hallowing_jump_scare_trap.py
jedgarpark: No, not yet, I was waiting on a decision on the tone A0 thing. Should I not have waited on that? Looks like that one was removed?

ladyada: i dont know how you saved your files but please commit it and push so we can wrap this up

jedgarpark: I just commited and pushed.

ladyada:
ok @jedgarpark for future code reviews, @TheKitty will help do stylistic review so they dont end up taking multiple days. you can chat on wednesdays/thursdays to get your code all shiny and ready for publication!

jedgarpark:
great, will do.

PR #350, opened 2018-10-04
3 d printed daft punk helmet

mikeysklar:
ported to CircuitPython


mikeysklar:
@ladyada Thank you for the rebase link and noticing that I had a stale repo. I believe I was able to clean it up with only the five files releated to the Daft Punk Helmet port in the repo.

ladyada:
ok just to check one of the .ino files has "Fiery horns" at the top - is that recycled code from noe & pedro?

mikeysklar:
The Daft Punk Helmet uses two Trinket's. One for a front facing lighting effect (which is "Fiery Horns" no modification) and the second Trinket to driver a side lighting effect which is the newly ported code. I can modify the comments for the front facing "Fiery Horns" if it could be a source of confusion.

ladyada:
all good thanx

PR #358, opened 2018-10-11
Add CircuitPython code for CPX compass project

dastels:


CPX_Compass/code.py
```diff
@@ -0,0 +1,146 @@
...
+    raw_maxes[1] = -1000.0
+    calibrate(True)
+
+
+    print("Calibration results")
+    print("Update the corresponding lines near the top of the code\n")
+    print("raw_mins = [{0}, {1}]".format(raw_mins[0], raw_mins[1]))
+    print("raw_maxes = [{0}, {1}]".format(raw_maxes[0], raw_maxes[1]))
+    while True:
+        pass
```
tannewt: Why busy wait here?

TheKitty: Dave, check please

dastels: It’s the calibration mode: Calibrate, output the lines to update the code with, and stall. Once the code is updated with the calibration results this code isn’t run and the compass works as a compass. Starting with button A pressed to  drop into this mode as well to,generate new calibration lines.

dastels: Alternatively, is there a way to exit to the repl immediately?

tannewt: try `sys.exit()`. That's how python would do it and I think it works in CircuitPython too.

dastels:
Done, tested, pushed, passed.  Ready to merge.

I don't seem to be getting notifications from github... something to look into.

ladyada:
thank you! mergin' :)

PR #378, opened 2018-11-02
Munny lamp

jedgarpark:


Munny_Lamp/Munny_Lamp.ino
```diff
@@ -0,0 +1,217 @@
...
+  analogWrite(GREEN_LED, 0);
+  analogWrite(BLUE_LED, 255);  // startup color, waiting for BLE connection
+
+  if (! lis.begin(0x18)) {   // change this to 0x19 for alternative i2c address
+    Serial.println("Couldnt start LIS3DH");
+    while (1);
+  }
+
+  Serial.begin(115200);
+  Serial.println(F("Adafruit Bluefruit MUNNY Neopixel Color Picker"));
```
ladyada: fix

jedgarpark: Oh, is it to fix the name since it isn't the NeoPixel we're using? Got it, thanks.

ladyada: yah is not neopixels!

jedgarpark:
OK, fixed up that comment @ladyada

PR #381, opened 2018-11-05
Changes to remove temp global variable

TheKitty:
Maintainers, pull request for John Park's review, please don't merge yet.

Per https://3.basecamp.com/3732686/buckets/4383362/todos/1319808931#__recording_1402164150

John, I took the check to see if the acceleration was positive and put that check directly in the if statement. Then the code to set tilt is not needed and the variable declaration is unneeded. Please review & try and then if this PR or your edits are good, we can look to merge.

### DIFF
--- a/Munny_Lamp/Munny_Lamp.ino
+++ b/Munny_Lamp/Munny_Lamp.ino
@@ -24,8 +24,6 @@ int blue = 0;
 #include "BluefruitConfig.h"
 Adafruit_LIS3DH lis = Adafruit_LIS3DH();
 
-int tilt = 0;  // for testing tilt of accelerometer
-
 /*=========================================================================
     APPLICATION SETTINGS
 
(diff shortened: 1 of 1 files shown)


Munny_Lamp/Munny_Lamp.ino
```diff
@@ -200,14 +198,7 @@ void accelerometer_check() {
...
 
-  if (event.acceleration.y < 0) {
-    tilt = 1;
-  }
-  else if (event.acceleration.y >= 0) {
-    tilt = 0;
-  }
-
-  if (tilt==1) {
+  if (event.acceleration.y >= 0) {
```
ladyada: this should be "< 0" to match the existing code? maybe even "< 4" so if your table is wobbly it wont go off :D

jedgarpark: yes, "<0" is the correct direction based on the build as I made it and the guide photos. Standing upright Y is around 9, totally upside down is -9, so 0 works well for the tipping point. This also allows him to blink if you lay him on his back, or not blink if you lay him on his stomach, just given how big his noggin is :)

jedgarpark: Sorry, @TheKitty I'm not skilled enough in Github foo to be sure I'll do the right thing trying to clone your fork and push a change to it, if that's even the right way to do it. Could you please change line 201 to:
`if (event.acceleration.y < 0) {`

TheKitty: I changed it to event.acceleration.y < 0, we'll see what Travis says

PR #402, opened 2018-11-20
simple midi uart example for TrellisM4 in CircuitPython

jedgarpark:


kattni:
@ladyada Is this good to go from your end?

ladyada:
yep this is fine, eventually we'll have a pretty MIDI library but for note on/off this works just fine :)

PR #411, opened 2018-11-25
Starduino 8bit super mario tree topper

mikeysklar:


mikeysklar:
@ladyada - Starduino is ready for QA.

ladyada:
nice - i like how short the code is with fancyled :)

mikeysklar:
so much cleaner with FancyLED

PR #423, opened 2018-12-06
Playing sounds and using buttons with raspberry pi

mikeysklar:
Two new CircuitPython audio / button playback scripts have been added for Raspberry Pi. The originals RPi.GPIO examples will fail to pass the lynting process.


ladyada:
i rather like mpg123, is omxplayer better?

mikeysklar:
I'm a fan of mpg123 as well. However, Raspbian installs omxplayer by default now. It is still a command line based player, but it supports all the popular audio/video formats (it will launch a window for the videos). Since omxplayer uses the OpenMax (omx) hardware acceleration is runs really smooth. Are you sold on it? I'd like to update the older RPi.GPIO code to look more like the CP code and use omxplayer and fix some syntax stuff if that is alright.

mikeysklar:
Okay. I will do an omxplayer update for the original RPi.GPIO examples. I'd like to try and lint/convert the original example to py3 syntax to please travis. I think the changes will be minimal just dropping some ()'s and a changing the 'not' logic syntax.

mikeysklar:
@ladyada - much happier with the python3 compliant syntax and omxplayer for the old code. Please approve for tutorial update.

PR #462, opened 2019-01-05
Update effected guides to use the debouncer library

dastels:


dastels:
Effected guides have been updated to reflect the use of the library.

PR #465, opened 2019-01-08
Drive a 16x2 lcd with the raspberry pi

mikeysklar:


mikeysklar:
@ladyada - Please QA this RPI 16x2 LCD update, init script change to be systemd compatible and Fritzing files the 26 pin and 40 pin cobbler.

PR #492, opened 2019-02-03
Raspberry pi email notifier using leds 

mikeysklar:
@ladyada - please QA at your convenience. Verified this works on Pi v1, v3b+.


mikeysklar:
I had vim open and accidentally checked in swap files from those two other projects. I believe I removed those, but they are still in the git log. I apologize for the extra noise. I had done a git add from the wrong directory after linting.

ladyada:
please merge once it passes travis :)

PR #497, opened 2019-02-06
first commit of BLE Crickit Light Switch code

jedgarpark:


BLE_Crickit_Light_Switch/ble_crickit_light_switch.py
```diff
@@ -0,0 +1,63 @@
...
+    blue_led.value = True
+
+    if uart_server.in_waiting:
+        red_led.value = False
+        packet = Packet.from_stream(uart_server)
+
+        if isinstance(packet, ButtonPacket):
+            if packet.pressed:
+                red_led.value = True
+                if packet.button == '5' and angle != UP_ANGLE:
```
kattni: We've added code for identifying the buttons by name which is a little clearer than this. I would suggest using the updated code which would look like this: 
```suggestion
                if packet.button == ButtonPacket.UP and angle != UP_ANGLE:
```

BLE_Crickit_Light_Switch/ble_crickit_light_switch.py
```diff
@@ -0,0 +1,63 @@
...
+                    # The Up button was pressed.
+                    for a in range(angle, UP_ANGLE+1, 1):
+                        crickit.servo_1.angle = a
+                    angle = UP_ANGLE
+
+                elif packet.button == '6' and angle != DOWN_ANGLE:
```
kattni: Likewise, this would look like this:
```suggestion
                elif packet.button == ButtonPacket.DOWN and angle != DOWN_ANGLE:
```

kattni:
I apparently needed to refresh! Dan beat me to it. Apologies!

dhalbert:
NP! GMTA!

BLE_Crickit_Light_Switch/ble_crickit_light_switch.py
jedgarpark: cool, I changed it to this

BLE_Crickit_Light_Switch/ble_crickit_light_switch.py
jedgarpark: I made this change

PR #502, opened 2019-02-10
simplify ble_crickit_light_switch.py

dhalbert:
This simplifies the code, removing `advertising_now` and removing moving the server one degree at a time. That code was there from the CircusPython to move the servo more slowly and to call `sparkle()` as continuously as possible. It's not necessary here.

The Learn Guide will need a tiny update to remove the sentence saying the servo is moved one degree at a time.

### DIFF
--- a/BLE_Crickit_Light_Switch/ble_crickit_light_switch.py
+++ b/BLE_Crickit_Light_Switch/ble_crickit_light_switch.py
@@ -26,38 +26,29 @@ UP_ANGLE = 180
 NEUTRAL_ANGLE = 120
 DOWN_ANGLE = 80
 
-advertising_now = False
-angle = NEUTRAL_ANGLE
 crickit.servo_1.angle = NEUTRAL_ANGLE
 
 while True:
+    blue_led.value = False
     crickit.servo_1.angle = NEUTRAL_ANGLE
-    if not uart_server.connected:
-        if not advertising_now:
-            uart_server.start_advertising()
-            advertising_now = True
-            blue_led.value = False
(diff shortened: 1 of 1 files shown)


dhalbert:
Tested with a servo, and tested that it can be disconnected and reconnect.

jedgarpark:
This no longer functions as it did with the anticipation of the action and neutral centering.

dhalbert:
Sorry, I misunderstood how this worked. Pushed another commit so it now returns to neutral after a brief delay to make sure the servo has gone all the way to the up or down position. You may want to increase the delay if needed.

The main point of this was to remove the 1-degree stepping, which was just there for `sparkle()`. The `advertising_now` logic was also due to `sparkle()`. I'm trying to make sure the canonical examples are as simple as possible.

jedgarpark:
Hmm, this version seems to only make the servo twitch a little bit in one direction and nothing in the other.

jedgarpark:
Oh, and I totally agree that the example shouldn't have the left over remains of the sparkle stuff, just want to get it working the way I had it from a functional standpoint.

dhalbert:
Ugh, that version was not pushed properly: it was missing a `time.sleep()`. Sorry about wasting your time on this. Try the latest version now.

jedgarpark:
Haha, who is Charlie Brown in this scenario ... :)

jedgarpark:
That's version is better, but it doesn't start with the anticipation motion that I had in the original. Sorry to be a pain about this, but I really loved the way it moved.

PR #533, opened 2019-02-25
Update CircuitPython_I2S_Tone.py

papahabla:
* Fixed issue with sine wave calculation where changing the frequency did not have the correct result
* The DC offset would change based on the `tone_volume` var, now it is constant at `2**15-1`. 
* Added sample rate variable to make sine wave generation more parametric.


TheKitty:
The lint program is complaining line 14 is > 79 chars (111). Try to use a continuation line with the correct indent or comment marks to have lines <= 80 chars (yeah, the linter is picky).

papahabla:
thanks @TheKitty, I think I figured out how to get through PEP8 adding a continuation line. The original line was also too long.

kattni:
Verified that guide does not need to be updated.

PR #543, opened 2019-02-27
ValueAdd fix to twitter.py and stars.py

jedgarpark:


jedgarpark:
tested both of these on PyPortal, all good.

PR #546, opened 2019-03-04
Move into the guide repo

dastels:


PyPortal_AlarmClock/code.py
```diff
@@ -12,18 +12,19 @@
...
 from adafruit_pyportal import PyPortal
 from adafruit_bitmap_font import bitmap_font
 from adafruit_display_text.text_area import TextArea
 from digitalio import DigitalInOut, Direction, Pull
 import analogio
 import displayio
-from secrets import secrets
 
 # Set up where we'll be fetching data from
 DATA_SOURCE = 'http://api.openweathermap.org/data/2.5/weather?q='+secrets['weather_location']
```
jerryneedell: Have you considered using the City ID here rather than name? If there are duplicate city names in a country it is hard to get the right one. id or zip gives more control.
for me - I live in Durham , NH , USA " but City name "Durham,US" does not work.
using "id" instead of "q"  with my city ID "508618"  gets me the right report

PyPortal_AlarmClock/code.py
```diff
@@ -0,0 +1,487 @@
...
+            return
+
+        # check light level and adjust background & backlight
+        self.adjust_backlight_based_on_light()
+
+        # only query the online time once per hour (and on first run)
+        if (not self.refresh_time) or ((now - self.refresh_time) > 3600):
+            try:
+                pyportal.get_local_time(location=secrets['time_location'])
+                self.refresh_time = now
```
jerryneedell: could this use "timezone" instead of "time_location" I think timezone is more commonly used.

jerryneedell:
downloaded and tested -- works!  It was not immediately obvious how to use the touch screen updates and they are very slow, but the basics are working well! I got it to change screens and adjust the alarm

dastels:
jerry.. yes, once screen updating is optimized, it will be smoother.

PyPortal_AlarmClock/code.py
dastels: Done

ladyada:
@dastels no biggie but if youc an convert to 16-bit BMPs, they take less space (and render a tad faster :)

PR #561, opened 2019-03-12
Updating PyPortal Weather Station to display_text labels

brentru:
Updating to work with display_text v2.0.0 breaking. Tested on a PyPortal running CPY 4.0.0 beta

https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/releases/tag/2.0.0


ladyada:
thanks - you need to change locations too, right?

PR #565, opened 2019-03-14
fixed text position tweak values

jedgarpark:
i had left in some offset value calculations, for example (200+20) which is now just (220).


ladyada:
hiya, why's that?

ladyada:
oops sorry ok i understand what you mean!

PR #566, opened 2019-03-14
Add PyPortal alarm clock stuff

dastels:


dastels:
Reworking for latest libraries

PyPortal_Alarm_Clock/code.py
```diff
@@ -0,0 +1,506 @@
...
+                    pass
+                filename = "/icons/"+weather_icon_name+".bmp"
+                if filename:
+                    if self.icon_file:
+                        self.icon_file.close()
+                    self.icon_file = open(filename, "rb")
+                    icon = displayio.OnDiskBitmap(self.icon_file)
+                    icon_sprite = displayio.TileGrid(icon,
+                                                     pixel_shader=displayio.ColorConverter(),
+                                                     position=(0, 0))
```
jerryneedell: This will no loner work with latest version of displayio - instead of position it uses x, y

PyPortal_Alarm_Clock/code.py
```diff
@@ -0,0 +1,506 @@
...
+            pyportal.splash.append(ta)
+        pyportal.splash.append(self.weather_icon)
+        if snooze_time:
+            if self.snooze_file:
+                self.snooze_file.close()
+            self.snooze_file = open('/icons/zzz.bmp', "rb")
+            icon = displayio.OnDiskBitmap(self.snooze_file)
+            icon_sprite = displayio.TileGrid(icon,
+                                             pixel_shader=displayio.ColorConverter(),
+                                             position=(0, 0))
```
jerryneedell: same as above - position  now uses x,y

ladyada:
is this a dupe?

dastels:
I have no idea what happened here... but **this** is the correct PR.  The other has a different directory name even. So it should be easy enough to delete it.

PR #581, opened 2019-03-25
Pyportal Oblique Strategies project added

CollinCunningham:


PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,50 @@
...
+pyportal.set_text("loading ...") # display while user waits
+
+pyportal.preload_font() # speed things up by preloading font
+
+while True:
+    # show title on first load
+    if firstload:
+        pyportal.set_text("OBLIQUE STRATEGIES\nBrian Eno / Peter Schmidt")
+        firstload = False
+    touch = pyportal.touchscreen.touch_point
```
ladyada: please simply this to not assign `touch` and then check it - just check the call direction, and debounce so it doesnt continue the loop until they no longer are touching

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,50 @@
...
+                   )
+                   
+pyportal.set_text("loading ...") # display while user waits
+
+pyportal.preload_font() # speed things up by preloading font
+
+while True:
+    # show title on first load
+    if firstload:
+        pyportal.set_text("OBLIQUE STRATEGIES\nBrian Eno / Peter Schmidt")
```
ladyada: seems like this can go before the loop if it only happens once?

CollinCunningham: yup. done.

ladyada:
ok please lint!

CollinCunningham:
@ladyada strategies.py has a string w 134 chars. Travis objects to line lengths over 100 char, but inserting line break breaks syntax. So, I'm just gonna merge this. Lemme know if that's an issue.

ladyada:
heya collin please do not merge things that you wrote, especially ones that break travis, it doesnt fix the problem, someone else has to do it and meanwhile eveyrone gets broken travis errors.

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,44 @@
...
+PyPortal screen is pressed. See the original Oblique Strategies
+by Brian Eno & Peter Schmidt here: https://www.enoshop.co.uk/product/oblique-strategies
+"""
+import os
+import random
+import board
+from strategies import strategies
+from adafruit_pyportal import PyPortal
+
+cwd = os.getcwd() # the current working directory (where this file is)
```
ladyada: this doesn't do what you think, use the original cwd assignment from other guides

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,44 @@
...
+
+# create pyportal object w no data source (we'll feed it text later)
+pyportal = PyPortal(url = None,
+                    json_path = None,
+                    status_neopixel = board.NEOPIXEL,
+                    default_bg = None,
+                    text_font = cwd+"fonts/Arial-ItalicMT-17.bdf",
+                    text_position = (30, 100),
+                    text_color = 0xFFFFFF,
+                    text_wrap = wrap,
```
ladyada: you dont't actually use this, remove

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,44 @@
...
+                    text_maxlen = 180, # max text size for quote & author
```
ladyada: you don't actually use this, remove

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,44 @@
...
+cwd = os.getcwd() # the current working directory (where this file is)
+wrap = 35 # text wrap value
+
+# create pyportal object w no data source (we'll feed it text later)
+pyportal = PyPortal(url = None,
+                    json_path = None,
+                    status_neopixel = board.NEOPIXEL,
+                    default_bg = None,
+                    text_font = cwd+"fonts/Arial-ItalicMT-17.bdf",
+                    text_position = (30, 100),
```
ladyada: why not center the text on (30, 120)? is this on purpose?

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,44 @@
...
+                   )
+
+pyportal.set_text("loading ...") # display while user waits
+pyportal.preload_font() # speed things up by preloading font
+pyportal.set_text("OBLIQUE STRATEGIES\nBrian Eno / Peter Schmidt") # show title
+
+while True:
+    if pyportal.touchscreen.touch_point:
+        # get random string from array and wrap
+        index = random.randint(0, len(strategies)-1)
```
ladyada: try random.choice(strategies)

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,44 @@
...
+        strat = pyportal.wrap_nicely(strategies[index], wrap)
+        # convert wrap array into line breaks
+        outstring = ""
```
ladyada: use the "".join() we have in pyportal

PyPortal_ObliqueStrategies/oblique.py
```diff
@@ -0,0 +1,44 @@
...
+
+cwd = os.getcwd() # the current working directory (where this file is)
+wrap = 35 # text wrap value
+
+# create pyportal object w no data source (we'll feed it text later)
+pyportal = PyPortal(url = None,
+                    json_path = None,
+                    status_neopixel = board.NEOPIXEL,
+                    default_bg = None,
+                    text_font = cwd+"fonts/Arial-ItalicMT-17.bdf",
```
ladyada: want to use a larger font? the strings seem short

CollinCunningham: The string length varies a lot - longest one is 127 characters

PyPortal_ObliqueStrategies/oblique.py
CollinCunningham: Ya - 100 was from when I tested with a larger font. I'll set it to  30, 120 in the next commit.

PyPortal_ObliqueStrategies/oblique.py
ladyada: kk

PyPortal_ObliqueStrategies/oblique.py
CollinCunningham: lovely - will do.

PyPortal_ObliqueStrategies/oblique.py
CollinCunningham: Just tested - it could use a point or two larger and still fit. Are there specs for converting to bdf in fontforge? Or a repo for bdf fonts?

PR #582, opened 2019-03-25
changed percent to dollar sign for USD, tweaked text position

jedgarpark:


jedgarpark:
This is a simple text fix if anyone wants to have a quick look, thanks!

PR #583, opened 2019-03-26
Updating PyPortal outlet design

djecken:
PyPortal Outlet Update


djecken:
no idea why this failed. just updating cad files.

dhalbert:
Build statuses appear wrong. Should be OK after merging now that I fixed the previous pylint problem.

PR #595, opened 2019-04-03
Patch PyPortal WeatherStation

brentru:
Adding directory for fonts, rename to `code.py`


ladyada:
u can merge once passin'

PR #598, opened 2019-04-03
Raspberry pi power control

mikeysklar:


mikeysklar:
@ladyada - please QA at your convenience.

PR #602, opened 2019-04-04
Update openweather_graphics.py

jasonlshelton:
Added day and date on second line of display. Minor adjustment of text placement to account for PyPortal stand that came in adabox 11


ladyada:
thanks! travis is a little sad about some linting stuff, can you check it out & fix :)
https://travis-ci.com/adafruit/Adafruit_Learning_System_Guides/builds/107155109

PR #625, opened 2019-04-10
Update display and brightness code in Magic_Nine_Ball

dhalbert:
Was using `Sprite` instead of `TileGrid`.
Now use builtin `brightness` control.

I'm seeing flickering when the brightness gets near 1.0, but I think that's an issue unrelated to this code, and probably needs to be fixed elsewhere.


ladyada:
@jedgarpark please update guide to reflect these fixes

jedgarpark:
Got it, will do.

PR #648, opened 2019-04-22
3d files for keyblade

djecken:
New 3d files for keyblade


djecken:
@TheKitty travis passed the cad files here – please review thx :-)

PR #670, opened 2019-04-26
Add example feature

isaacwellish:
- Was able to add a feature to switch between a word's definition and an example sentence. 

- Only error I can't remove from pylint is:

`Access to a protected member _json_path of a client class (protected-access)`

### DIFF
--- a/PyPortal_Word_of_the_Day/code.py
+++ b/PyPortal_Word_of_the_Day/code.py
@@ -1,10 +1,10 @@
 """
 This example uses the Wordnik API to display Wordnik's Word of the Day.
-Each day a new word, it's part of speech, and definition
-will appear automatically on the display.
+Each day a new word, its part of speech, and definition
+will appear automatically on the display. Tap the screen to start
+as well as to switch between the word's definition and an example sentence.
 """
 
-import time
 import board
 from adafruit_pyportal import PyPortal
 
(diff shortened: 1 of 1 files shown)


PyPortal_Word_of_the_Day/code.py
```diff
@@ -53,9 +56,18 @@
...
-    try:
-        value = pyportal.fetch()
-        print("Response is", value)
-    except RuntimeError as e:
-        print("Some error occured, retrying! -", e)
-    time.sleep(10*60) # Update every 10 minutes
+    if pyportal.touchscreen.touch_point:
+        try:
+            #set the JSON path here to be able to change between definition and example
+            pyportal._json_path=(WORD_LOCATION,
```
brentru: I have no issues with accessing this path as it's not an exposed property and there's no other way of changing it.

* Before Line 62, add the following line: 
`# pylint: disable=protected-access`

PyPortal_Word_of_the_Day/code.py
```diff
@@ -20,7 +20,11 @@
 WORD_LOCATION = ['word']
```
brentru: This doesn't need to be predefined, just use `['word']` on Line 62.

PyPortal_Word_of_the_Day/code.py
isaacwellish: Got it!

PyPortal_Word_of_the_Day/code.py
isaacwellish: kk!

brentru:
@isaacwellish  tag me again when you're ready to re-review and merge

isaacwellish:
@brentru ready to go!

brentru:
LGTM, thanks!

PR #685, opened 2019-05-04
Raspberry Pi DS18B20 temperature sensing

mikeysklar:
Missing CircuitPython support for busio.OneWire on Raspberry Pi. 

Using straight Python3 with Raspbian's built-in 1-wire support. This code attached here has been tested and runs fine, but it is not using CircuitPython libraries.

@ladyada - Any chance that busio.OneWire will be supported on RPi soon?

ow_bus = OneWireBus(board.D4)
File "/usr/local/lib/python3.5/dist-packages/adafruit_onewire/bus.py", line 77, in __init__
**self._ow = busio.OneWire(pin)**
**AttributeError: module 'busio' has no attribute 'OneWire'**


ladyada:
it will likely never be supported because we dont have userland access to 1wire. the current system is the only system!

PR #686, opened 2019-05-07
Heat set rig files

djecken:
Files for heat set rig


djecken:
@TheKitty Howdy, when you have a moment, Please merge these in – Thanks :_)

PR #687, opened 2019-05-07
moving nasa_token to secrets.py

jedgarpark:
just to stay consistent w other examples and get the url + api key out of code.py


ladyada:
mimic other guides

jedgarpark:
never mind, we are aborting the mission and leaving this code/guide alone!

PR #699, opened 2019-05-10
UK Admiralty tide data

DigitalMakerCIC:
Files to use UK Admiralty data

### DIFF
--- a/PyPortal_Tides/admiralty_tides.py
+++ b/PyPortal_Tides/admiralty_tides.py
@@ -1,0 +1,105 @@
+import time
+import board
+from adafruit_pyportal import PyPortal
+from adafruit_bitmap_font import bitmap_font
+from adafruit_display_text.label import Label
+import json
+from secrets import secrets
+
+#--| USER CONFIG |--------------------------
+STATION_ID = "0245"   # tide location, find yours from admiralty website/
+HI_COLOR   = 0x00FF00    # high tide times color
+LO_COLOR   = 0x11FFFF    # low tide times color
+DATE_COLOR = 0xFFFFFF    # date and time color
+#-------------------------------------------
+
+# pylint: disable=line-too-long
+DATA_SOURCE = "https://admiraltyapi.azure-api.net/uktidalapi/api/V1/Stations/" + STATION_ID + "/TidalEvents?duration=2"
+DATA_LOCATION = []
+
+# determine the current working directory needed so we know where to find files
+cwd = ("/"+__file__).rsplit('/', 1)[0]
+pyportal = PyPortal(url=DATA_SOURCE,
+                    headers={"Ocp-Apim-Subscription-Key":secrets['Ocp-Apim-Subscription-Key']},
+                    json_path=DATA_LOCATION,
+                    status_neopixel=board.NEOPIXEL,
+                    default_bg=cwd+"/tides_bg.bmp")
+
+# Connect to the internet and get local time
+pyportal.get_local_time()
+
+# Setup tide times font
+tide_font = bitmap_font.load_font(cwd+"/fonts/cq-mono-30.bdf")
+tide_font.load_glyphs(b'1234567890:')
+
+# Setup date and time font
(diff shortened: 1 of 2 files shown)


TheKitty:
Please make these minor fixes so the code will pas the Travis CI lint 1.9.2 checks
image

PyPortal_Tides/admiralty_tides.py
```diff
@@ -0,0 +1,105 @@
+import time
+import board
+from adafruit_pyportal import PyPortal
+from adafruit_bitmap_font import bitmap_font
+from adafruit_display_text.label import Label
+import json
+from secrets import secrets
```
caternuson: Here's a boiler plate that other programs have used for the secrets.py import:
```python
# Get wifi details and more from a secrets.py file
try:
    from secrets import secrets
except ImportError:
    print("WiFi secrets are kept in secrets.py, please add them there!")
raise
```

Example:
https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/master/PyPortal_OpenWeather/openweather.py#L14

PyPortal_Tides/admiralty_tides.py
```diff
@@ -0,0 +1,105 @@
...
+# Add all the labels to the display
+for label in HI_LABELS + LO_LABELS + [DATE_LABEL]:
+    pyportal.splash.append(label)
+
+def get_tide_info():
+    """Fetch JSON tide time info and return it."""
+
+    # Get raw JSON data
+    raw_info = pyportal.fetch()
+    raw_info = json.loads(raw_info)
```
caternuson: Were you not able to work with the return from `pyportal.fetch()` directly for some reason?

PyPortal_Tides/admiralty_tides_graphical.py
```diff
@@ -0,0 +1,157 @@
+import time
+import board
+import displayio
+import json
+from secrets import secrets
```
caternuson: Same comment as above.

PyPortal_Tides/admiralty_tides_graphical.py
```diff
@@ -0,0 +1,157 @@
...
+    time_marker_bitmap[pixel] = 2
+time_marker = displayio.TileGrid(time_marker_bitmap, pixel_shader=palette, x=-MARK_SIZE, y=-MARK_SIZE)
+pyportal.splash.append(time_marker)
+
+def get_tide_data():
+    """Fetch JSON tide data and return parsed results in a list."""
+
+    # Get raw JSON data
+    raw_data = pyportal.fetch()
+    raw_data = json.loads(raw_data)
```
caternuson: Same comment as above.

PyPortal_Tides/admiralty_tides.py
DigitalMakerCIC: changes made

PyPortal_Tides/admiralty_tides.py
DigitalMakerCIC: No it returned an array, that led to an error

PyPortal_Tides/admiralty_tides_graphical.py
DigitalMakerCIC: changes made

PyPortal_Tides/admiralty_tides_graphical.py
DigitalMakerCIC: Returned a JSON array that caused an error. This resolved it

caternuson:
Very sorry. My bad. It looks like my boiler plate copied in with bad formatting. It stripped out the spaces in front of `raise` for some reason. That line needs to be indented to line up with the `print`.
Let me try again:
```python
# Get wifi details and more from a secrets.py file
try:
    from secrets import secrets
except ImportError:
    print("WiFi secrets are kept in secrets.py, please add them there!")
    raise
```

The other two complaints are trailing-white space:
image

DigitalMakerCIC:
Think that's fixed it and removed trailing space.

PR #701, opened 2019-05-16
New example for continuous servo

TheKitty:
New example for continuous servo


ladyada:
dya wanna also put into the circuitpy motor library?

TheKitty:
This was requested by Dan for the CircuitPython Essentials and it's mirrored to other guides

TheKitty:
It'll be the standard continuous servo example for non-crickit wherever needed

PR #720, opened 2019-06-06
Lint AdaVoice code after move

TheKitty:


TheKitty:
Added WaveHC library to the arduino-libinstall list in the Adafruit_Learning_System_Guides repo
(Adafruit source: https://github.com/adafruit/WaveHC)

PR #726, opened 2019-06-10
Crank USB hid code

djecken:
Crank USB hid code


ladyada:
how is this code different than the media dial from before?

djecken:
just the neopixel time code has been stripped out.

djecken:
also it's config'd for the wiring on ItsyBitsy M0

ladyada:
i think maybe best to just use the same code, leave neopixel unconnected, and at the top have some commented out lines for whichever they want to use, then?

djecken:
sure i can add that back. btw, this code was not in learn repo - only embed in the media dial guide.

ladyada:
ok then lets add that code in, and you can adjust it!

PR #741, opened 2019-06-29
Adding Tilt Demo using BNO055 and PCA9685 for MagPi

makermelissa:
This is for the MagPi article I'm writing. If you'd rather me place it somewhere else, that's fine. It was the best place I could think of though.


ladyada:
all good thanx!

PR #746, opened 2019-07-06
Refactor to remove duplicate code

dcbriccetti:
I offer this, in case others value DRY (don’t repeat yourself).


dcbriccetti:
Makes perfect sense! Let’s do that. Thanks.

PR #756, opened 2019-07-10
Large pi based thermometer clock

mikeysklar:


ladyada:
can merge once travis passes

PR #762, opened 2019-07-12
Update feedback servo with CP examples

caternuson:
CircuitPython code examples and Feather M4 Express based Fritzing diagram as part of guide update:
https://learn.adafruit.com/analog-feedback-servos/


ladyada:
yum! merge when ready :)

PR #765, opened 2019-07-15
Twitter API guide code

dastels:


Twitter_API/binascii.py
```diff
@@ -0,0 +1,99 @@
+
+PAD = '='
+
+table_a2b_base64 = [
```
ladyada: please convert to a bytestring :)

Twitter_API/binascii.py
```diff
@@ -4,36 +4,30 @@
...
-    -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1,
-    -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1,
-    -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1,
-    -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1,
-    -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1,
-    -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1, -1,-1,-1,-1,
-]
-#pylint:enable=bad-whitespace
+#pylint:disable=bad-whitespace,invalid-name,bad-continuation
+table_a2b_base64 = bytes([
```
ladyada: use the *actual* bytestring, type this into cpython to get the bytestring and paste it in

Twitter_API/code.py
```diff
@@ -0,0 +1,133 @@
...
+the secrets dictionary must contain 'ssid' and 'password' at a minimum""")
+    raise
+
+def halt_and_catch_fire(message, *args):
+    """Log a critical error and stall the system."""
+    logger.critical(message, *args)
+    while True:
+        pass
+
+def connect_esp():
```
ladyada: ? why do we need this again when its in pyportal library?

Twitter_API/code.py
```diff
@@ -0,0 +1,133 @@
...
+import busio
+from adafruit_pyportal import PyPortal
+from digitalio import DigitalInOut
+from adafruit_esp32spi import adafruit_esp32spi
+import adafruit_esp32spi.adafruit_esp32spi_requests as requests
+import adafruit_logging as logging
+
+username = 'codewisdom'
+
+logger = logging.getLogger('main')
```
ladyada: i know you like logger stuff but lets keep it like the other pyportal demos and stick to print() :)

Twitter_API/code.py
```diff
@@ -0,0 +1,133 @@
...
+bearer_token = get_bearer_token()
+logger.debug('Got it')
+
+headers = {'Authorization': 'Bearer ' + bearer_token}
+
+url = 'https://api.twitter.com/1.1/statuses/user_timeline.json?count=1&screen_name=' + username
+while True:
+    logger.debug('fetching latest tweet from @%s', username)
+    response = requests.get(url,
+                            headers=headers)
```
ladyada: if you need to pass in headers, i think there's a pyportal PR that adds support? lets try that?

dastels: I tried using just the PyPortal class, but had problems fetching the bearer token before the PyPortal ... ... I think I might have a way around that...

ladyada: ok im interested if there's a way around it. if not, lets 'skip' the pyportal object and just DIY the whole thing with wifimanager & displayio

dastels: OK... I saw no PRs for a set_headers so I just use `pyportal._headers = '...'`

ladyada: oki!

dastels: Not exactly pristine software design.. but hey... it works.  Maybe an optional headers param to fetch at some point?

ladyada: if you want to add that to pyportal now i would merge it and then you could use that, seems handy!

dastels: PR submitted to PyPortal adding get_headers

PR #778, opened 2019-07-23
PyPortal Wake-Up Light

isaacwellish:
Code may need some refactoring but is fully functional.


isaacwellish:
Added some updates!

PyPortal_Wakeup_Light/wake_up_light.py
```diff
@@ -0,0 +1,198 @@
...
+MIN_BRIGHTNESS = 0
+MAX_BRIGHTNESS = 0.85
+#initialize neopixel strip
+num_pixels = 30
+ORDER = neopixel.RGBW
+strip = neopixel.NeoPixel(board.D3, num_pixels, brightness=BRIGHTNESS,
+                          pixel_order=ORDER)
+strip.fill(0)
+# color of strip
+WHITE = (255, 255, 255, 255)
```
ladyada: use just the white portion (0, 0, 0, 255) or you can red/yellow cast it with (64, 0, 0, 255)

PyPortal_Wakeup_Light/wake_up_light.py
```diff
@@ -0,0 +1,198 @@
...
+            if current_day == 6:
+                input_wake_up_time = wake_up_times[0]
+            else:
+                if current_day == day:
+                    input_wake_up_time = wake_up_times[day+1]
+    input_wake_up_time_text = "Wake up at " + input_wake_up_time
+    wakeup_time_textarea.text = input_wake_up_time_text
+    return input_wake_up_time
+
+def subtract30min(time_before):
```
ladyada: you should be able to do `time` based math instead of this parsing code
https://forum.micropython.org/viewtopic.php?t=959#p5492

PyPortal_Wakeup_Light/wake_up_light.py
```diff
@@ -0,0 +1,198 @@
...
+            pyportal.get_local_time()
+            refresh_time = time.monotonic()
+        except RuntimeError as e:
+            print("Some error occured, retrying! -", e)
+            continue
+    time_str_text = displayTime()
+    print(time_str_text)
+    currentHour = time.localtime()
+    # if after 9am and before 9pm light is = 0.8
+    if currentHour[3] > 9 and currentHour[3] < 21:
```
ladyada: dont turn on backlight until wake up time

PyPortal_Wakeup_Light/wake_up_light.py
```diff
@@ -0,0 +1,198 @@
...
+    # only query the online time once per hour (and on first run)
+    if (not refresh_time) or (time.monotonic() - refresh_time) > 3600:
+        try:
+            print("Getting time from internet!")
+            pyportal.get_local_time()
+            refresh_time = time.monotonic()
+        except RuntimeError as e:
+            print("Some error occured, retrying! -", e)
+            continue
+    time_str_text = displayTime()
```
ladyada: use time objects and math, dont store/display the time with PM or whatever, unless its to the screen

isaacwellish: For this do you want the user to input their wake up times as 24 hour time values? Reason I did it this way is so user can enter in their wake up times in 12 hour system (AM and PM).

ladyada: if you want them to enter wakeup time with A/P then parse that time *once* at the beginning of the program. and immediately convert it to a time object. you definitely should not be parsing/unparsing strings of time. makes the code clunkier than it needs to be

isaacwellish: Ah that makes sense. Ok will add those changes!

PR #781, opened 2019-07-30
Pyportal wakeup light updates

isaacwellish:
- display what day the wake up time is associated with

- add condition to starting wake up light

- fix backlight bug

- misc other bugs fixed


isaacwellish:
Sorry had to add one more change to fix a bug that was popping up on boot!

PyPortal_Wakeup_Light/wake_up_light.py
```diff
@@ -83,6 +84,12 @@
 pyportal.splash.append(wakeup_time_textarea)
 pyportal.splash.append(light_on_time_textarea)
 
+try:
```
ladyada: put this in a loop

isaacwellish: Oh right. What kind of loop should I put it in? It's usually just in the `while True:` loop but I had to call `pyportal.get_local_time()` outside of the `while True:` loop  one time before populating an array with wake up time objects. Otherwise the wake up time objects were not right. Is there a way to call pyportal.get_local_time() without using a try/except block?

ladyada: visit me

isaacwellish:
Github said my branch was out of date and prompted me to update so I did and travis failed? I tried syncing my repo in terminal then pushing and now I'm getting `Updates were rejected because the remote contains work that you do not have locally.` Not quite sure what to do :(

PR #787, opened 2019-08-02
Update Raspberry_Pi_Sensing_Movement.py

hwestman:
Added the pull up resistor. Seems to have been forgotten?

### DIFF
--- a/Raspberry_Pi_Sensing_Movement/Raspberry_Pi_Sensing_Movement.py
+++ b/Raspberry_Pi_Sensing_Movement/Raspberry_Pi_Sensing_Movement.py
@@ -9,6 +9,7 @@ pir_sensor.direction = digitalio.Direction.INPUT
 # set up door sensor
 door_sensor = digitalio.DigitalInOut(board.D23)
 door_sensor.direction = digitalio.Direction.INPUT
+door_sensor.pull = digitalio.Pull.UP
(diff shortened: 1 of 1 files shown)


hwestman:
Okey, I don't understand how you would get reliable values from the magnet sensor without the pull up resistor. Mine keeps reporting false negatives with this code. Any pointers on that would be appreciated :)

ladyada:
@hwestman not sure whats up but you could use any 10K for your setup :)

hwestman:
@ladyada you are right, I could use a 10k. I don't think I need one though since the PI has a built in pull up resistor. 

I thought some more about this and concluded that @mikeysklar is right, a pull up resistor would break this demo.
My take on it is that the demo here is to demonstrate a magnet switch reporting when the door has been opened. The sensor always report `false` as long as the door is closed, it only reports fluctuating values (what i called false negatives) when the door is open. And that behaviour is required for this scenario in order to reset the sensor, so it's not continually reporting that the door has been opened.

Am I somewhat close @mikeysklar ?

mikeysklar:
@hwestman Yes, I totally agree. door_sensor.value normally reports false when the door is closed and upon opening it will report true and print to the console "DOOR ALARM". I believe this is the logic we want to encourage for most door sensor usage.

PR #789, opened 2019-08-07
fix for SPIFlash - also need to use SdFat instead of SD

ladyada:
@brentru


ladyada:
@brentru plz approve/merge if this is working

PR #797, opened 2019-08-14
Add PyPortal UV Index display code

dastels:


PyPortal_UV_Index/code.py
```diff
@@ -0,0 +1,119 @@
...
+        json_payload = pyportal.fetch()
+        raw_data = json.loads(json_payload)
+    except (ValueError, RuntimeError) as ex:
+        print('Error: ', ex)
+        if isinstance(ex, ValueError):
+            print('JSON:', json_payload)
+        print('Retrying in 10 minutes')
+        time.sleep(600)
+        continue
+    data = [{'hour': extract_hour(d['DATE_TIME']), 'value': int(d['UV_VALUE'])}
```
ladyada: i like listcomps but they're a lil advanced, maybe use some temp variables for these entries so its a little easier to understand?

ladyada:
yay we can merge once passing

dastels:
FYI I don't have write perms, so I can't merge.

PR #799, opened 2019-08-15
Matrix 7 segment led backpack raspberry pi

mikeysklar:
Updated guide examples from LED_Backpack library to use adafruit_ht16k33 . pylint was complaining about display.print() being invalid syntax (syntax-error). This seems to be the norm we use with the other ht16k33 examples.


ladyada:
ok lemme know when it passes

PR #812, opened 2019-08-23
Mario scrolling clouds animation

dastels:


CircuitPython_Scrolling_Clouds/code.py
```diff
@@ -0,0 +1,137 @@
+"""
+This test will initialize the display using displayio and draw a solid green
```
ladyada: fix

dastels: Fixed and standard copyright/license/etc boilerplate added.

PR #817, opened 2019-08-27
fix compilation for hallowing m4

ladyada:


PaintYourDragon:
Fixed in recent commit along with other stuff.

PR #823, opened 2019-08-30
Flying toaster code

dastels:


CircuitPython_Flying_Toasters/code.py
```diff
@@ -0,0 +1,179 @@
+"""
+Continuously scroll randomly generated Mario style toasters.
```
ladyada: not Mario its from "After Dark"

CircuitPython_Flying_Toasters/code.py
```diff
@@ -0,0 +1,179 @@
...
+Designed fr an ItsyBitsy M4 Express and a 1.3" 240x240 TFT
```
ladyada: for

PR #826, opened 2019-09-06
Matrix 7 segment led backpack merge

mikeysklar:
Merged code examples from original LED_Backpack and HT16K33. Deprecating LED Backpack Display on Raspberry Pi and BeagleBone Black guide in favor of Matrix and 7-Segement LED Backpack with the Raspberry Pi guide.


ladyada:
ok we can merge once it passes!

PR #840, opened 2019-09-13
Add simple screen blanking for PiHole stats

caternuson:
Lots of people running this 24/7 and ruining their OLEDs. This adds a simple screen blanking so the display is only on for a short period of time.


ladyada:
please update the guide toooo! (im workin on a lil TFT version that wont have this issue)

PR #849, opened 2019-09-19
Add lots of yield()s, add user.cpp tab

PaintYourDragon:


ladyada:
will merge after CI passes

PR #865, opened 2019-09-25
Blank display when quitting via CTRL-C (Catch SIGINT)

Ackis:
Main change is to blank the display when quitting via ctrl-c. 

Did a few other clean up things, nothing major.


ladyada:
hiya please lint the code so it passes travis and we'll review it :)

PR #868, opened 2019-10-01
Configure PIR input pin (i.e. SENSE)

dastels:
SENSE input pin wasn't configured as INPUT so the PIR output was never actually read. Thus the display would blank after 5 seconds and never come back on.


ladyada:
OK! btw, in arduino, pins are input by default?

dastels:
Well... it wasn't set to INPUT and it didn't work. Add the line setting it to INPUT and it works as expected. Change in the latest Arduino framework?  I'm using 1.8.10

ladyada:
maybe changed ok! :)

PR #880, opened 2019-10-13
Reduced cap touch sampling to fix performance

jonsampson:
Frame rendering dropped precipitously on Hallowing M4 upon updating to adafruit/master user_touchneopixels.cpp (from 29 to 13 fps). I traced the issue to the Arcada capacitive touch button sampling. This patch simply reduces the number of times the buttons are read. This may make them a bit more finicky but that'd be my preference given the trade-off.

FPS still starts slow but eventually does reach 29 again.

### DIFF
--- a/M4_Eyes/user_touchneopixels.cpp
+++ b/M4_Eyes/user_touchneopixels.cpp
@@ -2,6 +2,7 @@
 
 #include "globals.h"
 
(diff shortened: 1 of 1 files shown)


jonsampson:
@djecken I confirmed capacitive touch on battery works as expected. :+1:

PR #916, opened 2019-11-03
improve touch response

tedsiy:


ladyada:
hiya this changes the behavior since now the last pin is prioritized so we wont merge - what are you trying to fix?

PR #919, opened 2019-11-05
Added snowflake

jedgarpark:
Added


caternuson:
yay. i meant to suggest just getting rid of the pin variables, like `tft_cs`, to clean up the boiler plate. glad to see that happened.

i might have also gotten rid of the magic numbers in the `randint`s and added color. but can save that for snowflakes 2 - tokyo snow drift.

PR #921, opened 2019-11-06
Scrolling clouds update

djecken:
Scrolling Clouds Update


djecken:
@TheKitty hiya, could you merge this in when you have a moment. it's an update to the scrolling clouds and flying toasters code. Thank you!

PR #925, opened 2019-11-08
Start a new repo for Dano's Walking Stick Guide

TheKitty:


caternuson:
Will `step_col` and `audio_file` ever be expanded? They both only contain a single entry and the index used is hardwired. So these could just become simple constants. Or, if there is a plan to allow more entries - how will they be added? This would lead to choice between list vs. tuple.

TheKitty:
@danowall see above

TheKitty:
> Will `step_col` and `audio_file` ever be expanded? They both only contain a single entry and the index used is hardwired. So these could just become simple constants. Or, if there is a plan to allow more entries - how will they be added? This would lead to choice between list vs. tuple.

Dano is hoping that as is it's a starting point for beginners to look at adding more than one music file, etc.

I suggest approving as-is if possible

caternuson:
OK. It's definitely functional as is.

PR #928, opened 2019-11-12
Add code for TFT Sidekick

caternuson:
Add code for TFT Sidekick


caternuson:
arduino checks appear to be broken? esp32 in particular.

ladyada:
yayyyy

PR #931, opened 2019-11-16
Add BLE CPB snow globe

caternuson:
Code for JP's BLE CPB Snowtastic Globery.

### DIFF
--- a/Snow_Globe_BLE_CPB/ble_snowglobe.py
+++ b/Snow_Globe_BLE_CPB/ble_snowglobe.py
@@ -1,0 +1,203 @@
+import time
+import random
+import board
+import busio
+import neopixel
+import adafruit_lis3dh
+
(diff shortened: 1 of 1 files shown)


caternuson:
more arduino build fail fun :(

caternuson:
what was the trick?

ladyada:
delete the cache - do you have permissions to do that?

caternuson:
yep. just not calibrated on how often that's needed. seems...often?

is there a typical pattern to look for?

ladyada:
if you're able to compile, its not a cache issues
if the core cant be found or a component is missing - thats because a new release came out and the cache is stale

ladyada:
wanna review the cpy GPS PR while you're here? :)

caternuson:
yep

PR #938, opened 2019-11-21
Update CPB Rover to 5.0.0-beta.0 BLE API

dhalbert:
Tested on a Crickit with motors, but with a CPB instead of a Feather. (I temporarily removed the status LED code to allow it to work with a CPB.) Seems to work fine.


dhalbert:
Changed wrong file. Closing this and will open another PR.

PR #949, opened 2019-11-30
update Luminary for updated BLE library and CP 5.0-beta.0 BLE

jerryneedell:
tested on CPB


dhalbert:
@firepixie can test on the actual hw.

PR #960, opened 2019-12-10
Removing color set as it is irrelevant.

kattni:
Color only matters on the first animation as you cannot get to the second or third animation without another CPB being connected, and at that point, the receiver is receiving color data that is setting the color of the animations.


ladyada:
needz a lint

PR #964, opened 2019-12-12
Add dreidel code

caternuson:
oh dreidel dreidel dreidel, i wrote this code today
oh dreidel dreidel dreidel, a pull request i shall make


ladyada:
a miracle happened here!

jerryneedell:
@caternuson FYI -- I tried run-in this on 2 CPBs wit the same result -- It starts up wit the greed with "gimmel" dislayed, then hangs -- no sound and no response to shake. Appears to he hung testing for the shake.

caternuson:
@jerryneedell Thanks for trying it out. So was that a keyboard interrupt that generated that traceback?

You know, I also thought the default was a bit stiff. I'll PR a quick rev that makes it settable and defaults to a lower threshold.

jerryneedell:
Yes, keyboard interrupt.
Thanks for making the change!

PR #965, opened 2019-12-12
Add dreidel shake threshold

caternuson:
Default shake threshold seems a bit stiff:
https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/964#issuecomment-564966052

This makes it changeable and defaults to a lower value.


ladyada:
thanks, it was a bit intensebefore

PR #972, opened 2019-12-16
Use Blocklist instead of blacklist

tannewt:
Inspired by https://github.com/kickstarter/rack-attack/pull/181


ladyada:
did not test code, but verified that the old variable does not exist

PR #974, opened 2019-12-19
Update DiscoTie to new BLE API

dhalbert:
A user in discord was trying to use the Disco_Tie code, and I found I'd left this off of the list of projects to update for the new API. Here is an updated version. I tested on the CPB, and added a setting to allow you to easily change whether to use external NeoPixels or those on the CPB (which the discord user wanted to do). Also added checking for malformed Bluefruit packets, which can happen if the user presses the keys too fast.

Once this is changed, the Guide needs to be changed to point to the newer versions of the libraries. See https://learn.adafruit.com/bluefruit-luminary-lanterns-with-capacitive-touch/software#install-libraries-3-6 as an example of some boilerplate about how to get the latest libraries, without needing direct links.


CollinCunningham:
@dhalbert Tested w `bundle-5.x-mpy-20191219`. Looks good but doesn't work unless ble is connected. I tried pulling that last block out of the `while ble.connected:` but lost all funcionality.

Also - that guide points to Adafruit_CircuitPython_BLE via a button that labelled as the Bundle.

Disco_Tie/code.py
```diff
@@ -221,12 +232,12 @@ def change_speed(mod, old_speed):
...
-    # Determine animation based on mode
-    if mode == 0:
-        peak = audio_meter(peak)
-    elif mode == 1:
-        rainbow_cycle(0.001)
-    elif mode == 2:
-        larsen(wait)
-    elif mode == 3:
-        solid(user_color)
+        # Determine animation based on mode
```
CollinCunningham: Ok - I got it working while disconnected by moving this block into an `animate(wait, peak)` function. 
Then calling `peak = animate(wait, peak)` at:

- last line of `while ble.connected:`
- before `pass` under `while not ble.connected:`

Is this a good solution?

Also - is there a better way for me to suggest a specific change without doing a PR on this branch? That seemed excessive.

CollinCunningham:
**FYI** I went ahead and fixed the library link in the lanterns guide.

Disco_Tie/code.py
CollinCunningham: let's merge this and I'll do anther PR for the changes I suggested.

PR #981, opened 2020-01-02
Provide USB HID devices directly rather than implicitly.

tannewt:
This makes it more obvious where in the examples to change in order
to switch to BLE HID.

This must be merged after https://github.com/adafruit/Adafruit_CircuitPython_HID/pull/43 is released.


tannewt:
@kattni please review. I'll update any surrounding text after it is merged.

kattni:
@tannewt The branch needs to be updated before merging.

tannewt:
Ok, guides are updated. If a CPX is used then I added a warning similar to:

> This example has been updated for version 4+ of the CircuitPython HID library. On the CircuitPlayground Express this library is built into CircuitPython. So, please use the latest version of CircuitPython as well. (At least 5.0.0-beta.3)

PR #982, opened 2020-01-03
Disco Tie - fix for animating while disconnected

CollinCunningham:
Moved animation calls to new `animate(pause, top)` function, called while both connected and not connected. This restores the original ability to view the customized animations after disconnecting from the app.


CollinCunningham:
@dhalbert Can you review this one?

PR #995, opened 2020-01-20
first commit CPB Volume BLE, added .python-version to .gitignore

jedgarpark:
.python-version is there because of Catalina and pylint shenanigans


dhalbert:
This looks really elegant! I am just awaiting the merging of this: https://github.com/adafruit/Adafruit_CircuitPython_BLE/pull/56 which makes the code sync up with the library.

PR #1001, opened 2020-01-22
BLE: Apple Media Service - Now Playing

makermelissa:
Here's the initial commit. It needs some fine tuning like shortening strings and font sizes, but is pretty close.


jedgarpark:
@makermelissa For some reason I'm having to reload a few times before it stops throwing an error on the font loading: `Traceback (most recent call last):
  File "code.py", line 84, in <module>
  File "code.py", line 49, in make_label
  File "code.py", line 43, in load_font
  File "adafruit_bitmap_font/bitmap_font.py", line 51, in load_font
OSError: [Errno 2] No such file/directory`

After a few reloads it works.

jedgarpark:
@makermelissa  This is looking good. I think we should use a smaller font for the track title, artist, and album text so more fits on screen. I also wonder if we can do text wrap to the next line and leave a blank line under each? See attached mockup.
amsTFTmockup06

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+from adafruit_gizmo import tft_gizmo
+from adafruit_display_text.label import Label
+from adafruit_display_shapes.rect import Rect
+from adafruit_bitmap_font import bitmap_font
+import adafruit_ble
+from adafruit_ble.advertising.standard import SolicitServicesAdvertisement
+from adafruit_ble_apple_media import AppleMediaService
+from adafruit_circuitplayground import cp
+
+BACKGROUND_COLOR = 0x444444 # Gray
```
jedgarpark: please change to this sage green `BACKGROUND_COLOR = 0x49523b`

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+TEXT_COLOR = 0xFF0000 # Red
+BORDER_COLOR = 0xAAAAAA # Light Gray
+STATUS_COLOR = BORDER_COLOR
+
+# PyLint can't find BLERadio for some reason so special case it here.
+radio = adafruit_ble.BLERadio() # pylint: disable=no-member
```
jedgarpark: add the radio name here `radio.name = "Now Playing Gizmo"`

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+    known_notifications = set()
+
+    # Draw the text fields
+    print("Loading Font Glyphs...")
+    title_label = make_label("None", 18, 30, TEXT_COLOR, font="/fonts/Arial-Bold-24.bdf")
+    artist_label = make_label("None", 18, 70, TEXT_COLOR, font="/fonts/Arial-22.bdf")
+    album_label = make_label("None", 18, 170, TEXT_COLOR, font="/fonts/Arial-22.bdf")
+    status_label = make_label("None", 80, 220, STATUS_COLOR, font="/fonts/Arial-16.bdf")
+    group.append(make_background(240, 240, BACKGROUND_COLOR))
+    border = Rect(10, 8, 200, 190, outline=BORDER_COLOR, stroke=1)
```
jedgarpark: Let's scale this out a bit and thicken the stroke `    border = Rect(4, 4, 232, 200, outline=BORDER_COLOR, stroke=2)
`

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+            pass
+
+        group.pop()
+    print("connected")
+
+    known_notifications = set()
+
+    # Draw the text fields
+    print("Loading Font Glyphs...")
+    title_label = make_label("None", 18, 30, TEXT_COLOR, font="/fonts/Arial-Bold-24.bdf")
```
jedgarpark: this can move left since we'll scale the rectangle up a bit, maybe 12 instead of 18, although this may change if we go with a smaller font.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+    artist_label = make_label("None", 18, 70, TEXT_COLOR, font="/fonts/Arial-22.bdf")
```
jedgarpark: spacing changes `artist_label = make_label("None", 12, 94, TEXT_COLOR, font="/fonts/Arial-22.bdf")`

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+    album_label = make_label("None", 18, 170, TEXT_COLOR, font="/fonts/Arial-22.bdf")
```
jedgarpark: spacing `    album_label = make_label("None", 12, 160, TEXT_COLOR, font="/fonts/Arial-22.bdf")
`

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+    status_label = make_label("None", 80, 220, STATUS_COLOR, font="/fonts/Arial-16.bdf")
```
jedgarpark: spacing     status_label = make_label("None", 80, 220, STATUS_COLOR, font="/fonts/Arial-16.bdf")

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+
+    while radio.connected:
+        try:
+            for connection in radio.connections:
+                if not connection.paired:
+                    connection.pair()
+                    print("paired")
+
+                ams = connection[AppleMediaService]
+                set_label(title_label, ams.title, 12)
```
jedgarpark: with more space we can print more characters, so 12 goes to 14, but should be even more with a smaller font

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+                set_label(album_label, ams.album, 13)
```
jedgarpark: increase max characters to fit w more space, smaller font

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
```diff
@@ -0,0 +1,130 @@
...
+                set_label(artist_label, ams.artist, 13)
```
jedgarpark: same max char note as above

makermelissa:
Though we can certainly use a smaller font, text wrapping is not currently a feature of the label. I'll have to look into seeing if adding that would be too difficult or not. I'll go ahead and adjust things and then looking text-wrapping. I'm not sure if that would involve making a new control based on the label or if it's as simple as adding some parameters to the existing label.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: Any suggestions on a font size? I was trying to find a balance between it showing lots of text and making it large like in your mockup. Like maybe 18 instead of 22?

jedgarpark:
Thanks @makermelissa OK, let's ignore text wrapping for now and it could be a nice-to-have at a later date.

jedgarpark:
@makermelissa since we aren't going to text wrap, we can keep the album name down at the bottom of the rectangle, too.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
jedgarpark: can leave album y coordinates at 170 since we aren't going to do text wrapping.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: done

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: done

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: done

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: I went with 18/16 and 12 is still fine.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: done, except with a smaller font, the y position around 80 looked better

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: done, except at 170 like you said later on.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: The 80 doesn't matter. This is automatically adjusted in set_status().

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: Increased to an appropriate amount for several songs. Since the fonts are not monospaced, this is more of a best guess kind of thing.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: same

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
makermelissa: same

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
jedgarpark: thanks, I'll check it out.

jedgarpark: the font sizes look good! I'm seeing 16 and bold 18, but not sure what you are referring to with "12 is still fine"?

jedgarpark: oh, sorry, I see, it's the position not the font size, nevermind :)

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
jedgarpark: I'd like to drop this even lower so it looks more intentional -- 184 works well, thanks.

CPB_AMS_Gizmo_BLE/cpb_ams_gizmo_ble.py
jedgarpark: Looks terrific!

jedgarpark:
@makermelissa the graphics and text work all looks fantastic now, thank you so much! A bug I'm seeing is that the code crashes if I the connection is lost w the phone. Either by walking out of range or turning off Bluetooth on phone. error is: `disconnected
  File "code.py", line 131, in <module>
AttributeError: 'Group' object has no attribute 'len'

Code done running. Waiting for reload.
Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.

Press any key to enter the REPL. Use CTRL-D to reload.`

makermelissa:
Ok cool. I'll look into that bug and adjust the 170 to 184.

PR #1002, opened 2020-01-24
Playing Now Gizmo better stability

makermelissa:
This change has stability improvements. The previous code would occasionally error out after running several hours. I moved the error handling to a smaller block and am using continue instead of pass and have not had it error out since.


ladyada:
@jedgarpark  please review

jedgarpark:
Oh, @makermelissa I just realized, this is an issue with the library, not your code: https://github.com/adafruit/Adafruit_CircuitPython_BLE_Apple_Media/issues/3
So, other than that, the code seems good to me.

makermelissa:
Ok, cool. Thanks for checking.

ladyada:
@makermelissa @jedgarpark please approve thru github so it can be merged

PR #1007, opened 2020-01-28
Updating openweather_graphics.py

BlitzCityDIY:
edit for update_date() function


djecken:
tested new update, running nicely 👍

PR #1012, opened 2020-01-31
first commit BLE heart rate trainer code

jedgarpark:


BLE_Heart_Rate_Trainer/ble_heart_rate_trainer.py
```diff
@@ -0,0 +1,167 @@
...
+# Start with a fresh connection.
+if ble.connected:
+    display_SCAN()
+    display_bLE()
+    time.sleep(1)
+
+    for connection in ble.connections:
+        if HeartRateService in connection:
+            connection.disconnect()
+        break
```
dhalbert: I think that these lines may no longer be needed with the fixed UF2 I sent you. You could try without.

BLE_Heart_Rate_Trainer/ble_heart_rate_trainer.py
```diff
@@ -0,0 +1,167 @@
...
+    display_A.set_digit_raw(2, 0b01110111)
+    display_A.set_digit_raw(3, 0b00110111)
+
+
+def display_bLE():
+    display_B.fill(0)
+    display_B.set_digit_raw(0, 0b00000000)
+    display_B.set_digit_raw(1, 0b01111100)
+    display_B.set_digit_raw(2, 0b00111000)
+    display_B.set_digit_raw(3, 0b01111001)
```
dhalbert: These make the main program easier to read and you could add routines like `display_dashes()` etc. for the low-level 7-segement calls below, even if they are used just once.

dhalbert:
This looks really good! The changes are cosmetic except maybe for removing the possibly unnecessary disconnect code.

BLE_Heart_Rate_Trainer/ble_heart_rate_trainer.py
jedgarpark: I'll give that a try, thanks.

jedgarpark: Yes, works without, I'll remove those lines.

BLE_Heart_Rate_Trainer/ble_heart_rate_trainer.py
jedgarpark: Will do, thanks.

ladyada: @makermelissa do you wanna make an optional 7-seg font table?

dhalbert:
I will not merge in case anyone else wants to weigh in before. But fine to merge.

ladyada:
can adapt later!

PR #1020, opened 2020-02-06
[WIP] Pyloton

dherrada:
~Code's currently pretty unreadable. I should have it decently legible sometime today or tomorrow.~ I'll be tagging everyone currently on the basecamp thread.


dherrada:
Should this repository be on Actions?

ladyada:
nope

dherrada:
I haven't tested the last few commits as they were aimed at readability and not functionality. I will test them sometime this afternoon/night.

dherrada:
Pyloton is resonably ready for review. I wouldn't call it finished, but I wanted to get some review before moving forward any more.

siddacious:
@dherrada I'm testing with no sensors and it looks like you're giving some status information at the top of the screen but it's getting cut off:
IMG_20200210_162957__01

siddacious:
it also appears to be writing over the same text without clearing the area first

dherrada:
@siddacious What version of firmware do you have on the board? I haven't been able to replicate any of the issues you and @jedgarpark have been having.

dherrada:
@siddacious Oh, are you running it with all of the sensors disabled? As in pyloton(heart=False, cadence=False...)

dherrada:
Ok. I've added AMS to this. Feel free to test, but I don't see a big reason to until we find out why we're getting such drastically different results.

siddacious:
@dherrada I was testing with the stock code from your fork with no sensors. I used the latest build available on S3

dherrada:
@siddacious Try 5.0.0-beta.5. That's what I'm using and I've found it to be much more stable. Will test with most recent s3 to see if I can replicate your issues.

siddacious:
@dherrada please merge master with your branch

dherrada:
@siddacious I believe commit 782489f. Lemme know if I misunderstood what you were asking for. Thanks!

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+start = time()
+hr_connection = None
+speed_cad_connections = []
+radio = None
+while True:
+    if heart:
+        if not hr_connection:
+            print("Running hr_connection")
```
dhalbert: If you want to leave these print statements in, make them more understandable, like "Connecting to heart rate monitor", etc.

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,59 @@
...
+            hr_connection = pyloton.heart_connect()
+            ble.stop_scan()
+    if speed or cad:
+        if not speed_cad_connections:
+            print("Running speed_cad_connection")
```
dhalbert: "connections" plural

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,59 @@
...
+ble = adafruit_ble.BLERadio()    # pylint: disable=no-member
+
+display = board.DISPLAY
+
+
+heart = True
+speed = True
+cad = True
+ams = True
+debug = False
```
dhalbert: I would make these all caps because they are "constants" used to switch things on and off.

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,59 @@
...
+        if not hr_connection:
+            print("Running hr_connection")
+            hr_connection = pyloton.heart_connect()
+            ble.stop_scan()
+    if speed or cad:
+        if not speed_cad_connections:
+            print("Running speed_cad_connection")
+            speed_cad_connections = pyloton.speed_cad_connect()
+
+    if time()-start >= 45:
```
dhalbert: Make a constant  "CONNECTION_TIMEOUT" or something like that.

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,59 @@
...
+        pyloton.timeout()
+        break
+    # Stop scanning whether or not we are connected.
+    ble.stop_scan()
+
+    # You may need to remove some parts of the following 2 lines depending on what
+    # devices you are using.
```
dhalbert: Need to explicate this more.

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,59 @@
...
+    if hr_connection and speed_cad_connections and ams:
+        if hr_connection.connected and speed_cad_connections[0].connected and ams.connected:
+            pyloton.setup_display()
+            break
+
+# You may need to remove some parts of the following line depending on what devices you are using.
```
dhalbert: ditto

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,563 @@
...
+        Connect to an Apple device using the ble_apple_media library
+        """
+        self._status_update("AppleMediaService: Connect your phone now")
+        radio = adafruit_ble.BLERadio()
+        a = SolicitServicesAdvertisement()
+        a.solicited_services.append(AppleMediaService)
+        radio.start_advertising(a)
+
+        while not radio.connected:
+            pass
```
dhalbert: This will loop forever waiting to connect. Do you want to have a timeout here?

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,563 @@
...
+    splash = displayio.Group(max_size=25)
+    loading_group = displayio.Group()
+
+    YELLOW = 0xFCFF00
+    PURPLE = 0x64337E
+    WHITE = 0xFFFFFF
+
+    start = time.time()
+
+    track_artist = True
```
dhalbert: Many of these are class variables, but they are used like instance variables. I'd try to be consistent and assign them in `__init__()`, or else don't make this a class (but a class is clearer).

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,563 @@
...
+    def read_s_and_c(self):
+        """
+        Reads data from the speed and cadence sensor
+        """
+        speed = self._previous_speed
+        cadence = self._previous_cadence
+        for conn, svc in zip(self.cyc_connections, self.cyc_services):
+            if not conn.connected:
+                speed = cadence = 0
+                continue
```
dhalbert: This will loop forever when there are no connections. Maybe return instead so caller can take note of connection loss?

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,563 @@
...
+            self._previous_revolutions = values.cumulative_wheel_revolutions
+            self._i = 0
+        else:
+            self._i += 1
+            if self._i >= 3:
+                speed = 0
+        self._previous_wheel = values.last_wheel_event_time
+
+
+    def _cad_helper(self, values):
```
dhalbert: Would it make sense to call this `_compute_cadence_and_speed` or similar?

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,563 @@
...
+            # We then mutiply it by the wheel's circumference and convert it to mph
+            speed = round((rpm * self.circumference) * (60/63360), 1)
+            if speed < 0:
+                speed = self._previous_speed
+            self._previous_speed = speed
+            self._previous_revolutions = values.cumulative_wheel_revolutions
+            self._i = 0
+        else:
+            self._i += 1
+            if self._i >= 3:
```
dhalbert: can `._i` and `._j` have better names or more documentation?

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,563 @@
...
+            text_group.append(status)
+
+
+
+        if len(self.loading_group) < 4:
+            self.loading_group.append(text_group)
+        else:
+            self.loading_group[3] = text_group
+
+        self.display.show(self.loading_group)
```
dhalbert: Could you create the labels and groups just once and reuse them?

CircuitPython_Pyloton/pyloton.py
dhalbert: It looks like am wrong about this. I thought it was a `while True` for some reason.

dhalbert: But if only one of two sensors is not connected, we'll lose data for both? Is that OK?

CircuitPython_Pyloton/code.py
dherrada: @dhalbert How does this sound?
``` 
# You may need to remove some parts of the following line depending on what devices you are using.
# For example, remove hr_connection and hr_connection.connected if you aren't using a heart rate
# monitor. Do the same for other sensors.
```

dhalbert: You could do this in code by testing whether `heart` is `True` rather than by asking the user to edit the code.

CircuitPython_Pyloton/pyloton.py
dherrada: Yeah, I think so

CircuitPython_Pyloton/pyloton.py
dherrada: Personally, I think that's alright. This is just put in there because a sensor may disconnect after the main loop in code.py checks if it's still connected, and that can cause read_s_and_c to fail and give an ugly error message. This just sets both to zero and then the main loop should be broken out of on the next run.

dhalbert: I understand now. Resolving this convo.

CircuitPython_Pyloton/pyloton.py
dherrada: @dhalbert The reason I was doing it this way was that sometimes it would be one line and sometimes it would be two. But, I do agree that doing it your way would be cleaner and probably clean things up a bit.

dherrada:
@dhalbert I made your requested changes. I haven't tested yet as I've got to head out for a bit, but when I get back I'll test it.

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,64 @@
...
+# For example, remove hr_connection and hr_connection.connected if you aren't using a heart rate
+# monitor. Do the same for other sensors you aren't using.
+while hr_connection.connected and speed_cad_connections[0].connected and ams.connected:
+    pyloton.update_display()
+    pyloton.ams_remote()
```
dhalbert: So I'm suggesting to do something like this, which will check at runtime which devices are enabled:
```python
if ((not HEART or (hr_connection and hr_connection.connected)) and
    ((not SPEED or not CAD) or (speed_cad_connections and speed_cad_connections[0].connected)) and
    (not AMS or (ams and ams.connected))):
    break

pyloton.setup_display()

while ((not HEART or hr_connection.connected) and
    ((not SPEED or not CAD) or speed_cad_connections[0].connected) and
    (not AMS or ams.connected)):
    pyloton.update_display()
    pyloton.ams_remote()
```

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,571 @@
...
+        if len(message) > 25:
+            status.text = message[:25]
+            status1.text = message[25:]
+
+            text_group.append(status)
+            text_group.append(status1)
+        else:
+            status.text = message
+            status1.text = ''
+            text_group.append(status)
```
dhalbert: `status` and `status1` won't have a value if the `if len(...` is False. You need to make them instance vars to remember them.

Testing for `len(self.loading_group) == 3` is "brittle", because the `3` might change. You could test for `status` and/or `status` being non-`None` instead.

CircuitPython_Pyloton/code.py
dherrada: Ooh, that’s a good idea

CircuitPython_Pyloton/pyloton.py
dherrada: Will do

CircuitPython_Pyloton/code.py
dhalbert: maybe it's `not SPEED and not CAD`, since you want to skip checking for speed or cad connections only if both are off

dherrada:
@dhalbert I have tested this version and it does work

dherrada:
@dhalbert do you think you might be able to take another look at this when you got a sec? Thanks!

dherrada:
@dhalbert sounds good. I’ll test this again at like 1:30 today and then let you know if i think it’s ready

dherrada:
@dhalbert Ok, I've tested and I'm confident it's working well. Waiting on others to test.

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,63 @@
+from time import time
+import adafruit_ble
+import board
+import pyloton
+
+# PyLint can't find BLERadio for some reason so special case it here.
```
ladyada: remove this line

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,63 @@
...
+ble = adafruit_ble.BLERadio()    # pylint: disable=no-member
+
+display = board.DISPLAY
```
ladyada: remove this line

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,63 @@
...
+CONNECTION_TIMEOUT = 45
+
+HEART = True
+SPEED = True
+CADENCE = True
+AMS = True
+DEBUG = False
+
+# 84.229 is wheel circumference (700x23 in my case)
+pyloton = pyloton.Pyloton(ble, display, 84.229, HEART, SPEED, CADENCE, AMS, DEBUG)
```
ladyada: pass `board.DISPLAY`

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,63 @@
...
+
+pyloton.show_splash()
+
+ams = pyloton.ams_connect()
+
+
+start = time()
+hr_connection = None
+speed_cadence_connections = []
+radio = None
```
ladyada: is this variable used anywhere?

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+        Displays status updates
+        """
+        if self.debug:
+            print(message)
+            return
+
+        if self.text_group not in self.loading_group:
+            self.loading_group.append(self.text_group)
+
+        if len(message) > 25:
```
ladyada: clean this up the way you did the other long texts

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+            self.status.text = message[:25]
+            self.status1.text = message[25:50]
+
+        else:
+            self.status.text = message
+            self.status1.text = ''
+
+        #self.display.show(self.loading_group)
```
ladyada: if not needed, remove

CircuitPython_Pyloton/code.py
dherrada: Yeah, it's used in line 38 (will be 35 after next push) so there isn't a NameError.

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+
+        self._status_update("AppleMediaService: Connected")
+
+        #known_notifications = set()
+
+        for connection in radio.connections:
+            if not connection.paired:
+                connection.pair()
+                self._status_update("AppleMediaService: Paired")
+
```
ladyada: remove space

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+        a = SolicitServicesAdvertisement()
+        a.solicited_services.append(AppleMediaService)
+        radio.start_advertising(a)
+
+        while not radio.connected and not self._has_timed_out(start, timeout):
+            pass
+
+        self._status_update("AppleMediaService: Connected")
+
+        #known_notifications = set()
```
ladyada: remve if not used

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+        if not advs:
+            # Nothing found. Go back and keep looking.
+            return []
+
+        # Connect to all available CSC sensors.
+        self.cyc_connections = []
+        for adv in advs.values():
+            self.cyc_connections.append(self.ble.connect(adv))
+            self._status_update("Speed and Cadence: Connected {}".format(len(self.cyc_connections)))
+
```
ladyada: remove space

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+
+        self.cyc_services = []
+        for conn in self.cyc_connections:
+
```
ladyada: remove space

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+                    if self._speed_failed > 3:
+                        speed = 0
+                continue
+
+            if not values.last_wheel_event_time:
+                continue
+
+            speed = self._compute_speed(values, speed)
+
+
```
ladyada: remove space

CircuitPython_Pyloton/pyloton.py
```diff
@@ -0,0 +1,584 @@
...
+            ams_sprite = self.icon_maker(3, 2, self._ams_y - 20)
+            sprites.append(ams_sprite)
+
+        self.splash.append(sprites)
+
+        self.display.show(self.splash)
+        while self.loading_group:
+            self.loading_group.pop()
+
+
```
ladyada: remove space

CircuitPython_Pyloton/code.py
ladyada: i was referring to 'radio = None'

CircuitPython_Pyloton/code.py
```diff
@@ -0,0 +1,60 @@
...
+            print("Attempting to connect to speed and cadence monitors")
+            speed_cadence_connections = pyloton.speed_cadence_connect()
+
+    if time()-start >= CONNECTION_TIMEOUT:
+        pyloton.timeout()
+        break
+    # Stop scanning whether or not we are connected.
+    ble.stop_scan()
+
+    if ((not HEART or (hr_connection and hr_connection.connected)) and #pylint: disable=too-many-boolean-expressions
```
ladyada: put the pylint disable on the previous line

CircuitPython_Pyloton/code.py
dherrada: Ah, ok

ladyada:
will merge on CI pass

PR #1036, opened 2020-02-26
Clue dice

dastels:
New PR with errant M4eyes file removed.
Comments from previous PR incorporated.


CLUE_Dice_Roller/clue_dice_roller.py
```diff
@@ -0,0 +1,108 @@
...
+display
+mode = SELECTING
+update_display(number_of_dice, side_selection)
+
+while True:
+    button_a.update()
+    button_b.update()
+
+    if mode == SELECTING:
+        if button_a.rose:
```
ladyada: i think people expect it to change on falling (pressed) not rising?

dastels: The clue button properties return True when pressed.  As does the CPB buttons.  Both those surprised me for just that reason.

ladyada:
ok nice! plz lint :)

dastels:
Yeah.. linting... it's been a while :)

dglaude:
I believe d10 is missing from your code, but is advertised in the comment:

SIDES = [4, 6, 8, 12, 20, 100] =>SIDES = [4, 6, 8, 10, 12, 20, 100]

dastels:
Doah.. d10 is definitely missing.

PR #1038, opened 2020-02-27
init sketch

ladyada:
openweather based animated cloud

### DIFF
--- a/CircuitPython_WeatherCloud/code.py
+++ b/CircuitPython_WeatherCloud/code.py
@@ -1,0 +1,121 @@
+import time
+import board
+import busio
+import random
+from digitalio import DigitalInOut
+import neopixel
+from adafruit_esp32spi import adafruit_esp32spi
+from adafruit_esp32spi import adafruit_esp32spi_wifimanager
+import adafruit_requests as requests
+import adafruit_fancyled.adafruit_fancyled as fancy
+
+print("ESP32 Open Weather API demo")
+
+# Use cityname, country code where countrycode is ISO3166 format.
+# E.g. "New York, US" or "London, GB"
+LOCATION = "Manhattan, US"
+
+# Get wifi details and more from a secrets.py file
+try:
+    from secrets import secrets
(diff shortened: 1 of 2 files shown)


CircuitPython_WeatherCloud/code.py
```diff
@@ -0,0 +1,119 @@
...
+weather_refresh = None
+weather_type = None
+while True:
+    # only query the weather every 10 minutes (and on first run)
+    if (not weather_refresh) or (time.monotonic() - weather_refresh) > 600:
+        try:
+            response = wifi.get(DATA_SOURCE).json()
+            print("Response is", response)
+            weather_type = response['weather'][0]['main']
+            weather_type = 'Thunderstorm' # manually adjust weather type for testing!
```
caternuson: remove?

CircuitPython_WeatherCloud/code.py
```diff
@@ -0,0 +1,119 @@
...
+pal_offset = 0  # Positional offset into color palette to get it to 'spin'
+levels = (0.25, 0.3, 0.15)  # Color balance / brightness for gamma function
+raining = False
+thundering = False
+
+weather_refresh = None
+weather_type = None
+while True:
+    # only query the weather every 10 minutes (and on first run)
+    if (not weather_refresh) or (time.monotonic() - weather_refresh) > 600:
```
caternuson: Could you init `weather_refresh` to some negative number, like `-600` to make it run on first pass? Then this could just become:
```python
if time.monotonic() - weather_refresh > 600:
```

ladyada: i kinda prefer this cause you could 'force' refresh whenever you want :)

CircuitPython_WeatherCloud/code.py
ladyada: this is for another person to test so its on purpose :)

ladyada:
thank u!

PR #1048, opened 2020-03-07
Display CPU temperature in pihole stats script

ehershey:
The temperature command was set up but not executed or used for display


dhalbert (approved):
Looks ok.

PR #1055, opened 2020-03-14
Changed function call to property reference

axiomatic-io:
An update to the APDS9960 library uses a property to reference the proximity value rather than a function call


axiomatic-io:
Sorry forgot to tag @kattni

caternuson:
Just verifying that there's no other code to visit. Appears not.
```
Adafruit_Learning_System_Guides$ grep -i -r apds9960 | grep proximity
Adafruit_Feather_Sense/feather_sense_sensor_demo.py:apds9960.enable_proximity = True
Adafruit_Feather_Sense/feather_sense_sensor_demo.py:    print("Proximity:", apds9960.proximity())
```

PR #1060, opened 2020-03-27
waterfall: adjust to changed name of function

jepler:
Since the time I wrote this code, the function was moved to "extras" and renamed.  Adjust.


jepler:
Testing performed: Loaded waterfall.py on clue with 5.1.0-rc.0 and ran it

PR #1064, opened 2020-03-31
first commit

jedgarpark:


makermelissa:
You could speed this up a bit by changing the preload strings that never change to just the string value.
For instance, for title_font, instead of "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz", you could just use "Hand Wash" and it will only load those characters.

jedgarpark:
@makermelissa cool, I'll do that.

CLUE_Hand_Wash_Timer/clue_hand_wash_timer.py
```diff
@@ -0,0 +1,88 @@
+"""
+Start a 20 second hand washing timer via proximity sensor.
+Countdown the seconds with text and beeps.
+Display a bitmaps for waiting and washing modes.
+"""
+
+import time
+import board
```
makermelissa: You could avoid importing board and just use `clue.display`

jedgarpark: that changes the display calls throughout right?

jedgarpark: Oh, nevermind, I see what you mean. Got it, will do.

jedgarpark: hmm, seems board needs to be imported for buzzer to work, unless there's a clue equivalent to board.SPEAKER

jedgarpark: I'll switch to `clue.display` since it is more consistent with the other display stuff here, but leaving board import in because it's needed for the way buzzer (pulseio) is being used without a duration. Seems` clue.playTone()` would require a duration and possibly add an unwanted delay to the countdown.

makermelissa: Oh yeah, good call. Go ahead and leave it in.

PR #1071, opened 2020-04-08
change read_region()

liltux:
add time.sleep(.3) delays and break(s) to exit loop upon touch.  (my opinion) more responsive game.  do not have to wait for timer to expire (30 seconds) between touches.


liltux:
Added a commit and change for cpx_hunter.py file.  debug was removed from adafruit_irremote.py in: https://github.com/adafruit/Adafruit_CircuitPython_IRRemote/pull/27.  The debug=False was causing code to fail.

caternuson:
@ladyada Sorry, just coming across this. Both changes look good. Tested first one and it works. Second one looks like simple update to match a breaking change.

ladyada:
np merge once it passes :)

caternuson:
@liltux Thanks for the updates!

liltux:
@everyone_adafruit you guys are awesome,  I love the circuit playground boards.  I just came across some things while working on projects to do with my kiddos.

PR #1072, opened 2020-04-09
PyPortal User Interface - fallback if we don't have the temperature sensor

FoamyGuy:
This change will allow the example to work on the Titano and Pynt devices which do not have the temperature sensor.


FoamyGuy:
Neat! I did not know about this. Thank you

ladyada:
awesome!

PR #1075, opened 2020-04-14
first commit powerglove ble midi

jedgarpark:


PowerGlove_BLE_MIDI/powerglove_ble_midi.py
```diff
@@ -0,0 +1,126 @@
...
+        )
+        time.sleep(0.2)
+
+    else:
+        print("Waiting for connection")
+        while not ble.connected:
+            pass
+        print("Connected")
+        while ble.connected:
+
```
tannewt, line 86: Kinda weird to have a blank line as the first after the previous ends in :

jedgarpark: right on, will fix.

PR #1076, opened 2020-04-14
Demo files for CircuitPython Protomatter guide

jepler:


CircuitPython_Protomatter/Protomatter.py
```diff
@@ -0,0 +1,19 @@
...
+                        native_frames_per_second=60):
+    height = len(rgb_pins) // 3 * 2 ** len(addr_pins)
+    pm = _protomatter.Protomatter(
+        width=width, height=height, bit_depth=bit_depth,
+        rgb_pins=rgb_pins, addr_pins=addr_pins, clock_pin=clock_pin,
+        latch_pin=latch_pin, output_enable_pin=output_enable_pin,
+        doublebuffer=doublebuffer, framebuffer=framebuffer,
+    )
+    return framebufferio.FramebufferDisplay(
+        pm, width=width, height=height,
```
tannewt, line 16: Could FramebufferDisplay just inherit width and height from the pm? That would simplify this constructor and remove the need for this helper file.

jepler: That would be a great improvement, but needs changes in core first.

jepler:
I will open a fresh PR once the protomatter PR has actually landed in CircuitPython

PR #1077, opened 2020-04-15
CircuitPython Protomatter: alternate version

jepler:
This addresses the review comment by Scott (and is a nice simplification of the code), but requires that adafruit/circuitpython#2774 be merged first.  If this PR is taken, then #1076 should be closed.


jepler:
I will open a fresh PR once the protomatter PR has actually landed in CircuitPython

PR #1080, opened 2020-04-19
CircuitPython RGBMatrix demo programs

jepler:
This depends on adafruit/circuitpython#2775 which has not yet been merged.

Compared to #1077 this follows the renamed module and also adds a Conway's Life demo.


ladyada:
i hand tested each one - works great!

PR #1084, opened 2020-04-22
update benchmark.py

v923z:
This PR gains a factor of 3 in speed, when calculating the standard deviation of a set of data with `ulab`. 

Times measured on a pyboard.v.1.1, `ulab` v.0.42.0 are as follows:

```
Computing the RMS value of 100 numbers
in python
execution time:  2846  us

in ulab, with some implementation in python
execution time:  175  us

in ulab only, with ndarray
execution time:  78  us

in ulab only, with list
execution time:  245  us
```


ladyada:
thanks - can you run pylint on the example and update the PR :)
https://travis-ci.com/github/adafruit/Adafruit_Learning_System_Guides/builds/161374439#L1036

ulab_Crunch_Numbers_Fast/benchmark.py
```diff
@@ -1,11 +1,21 @@
-import time
+import utime
```
jepler, line 1: This is not correct for CircuitPython.  Please check that the code runs on CircuitPython, as that's what the Guide covers.

ulab_Crunch_Numbers_Fast/benchmark.py
```diff
@@ -1,11 +1,21 @@
...
 import math
 import ulab
 import ulab.numerical
 
+def timeit(f, *args, **kwargs):
+    func_name = str(f).split(' ')[1]
+    def new_func(*args, **kwargs):
+        t = utime.ticks_us()
+        result = f(*args, **kwargs)
+        print('execution time: ', utime.ticks_diff(utime.ticks_us(), t), ' us')
```
jepler, line 11: ticks_diff and ticks_us do not exist in CircuitPython.  We have time.monotonic_ns on all our platforms.  Naturally you can divide by 1000 to convert to us.

ulab_Crunch_Numbers_Fast/benchmark.py
```diff
@@ -1,11 +1,21 @@
...
+        return result
+    return new_func
+
 def mean(values):
     return sum(values) / len(values)
 
+@timeit
```
jepler, line 18: I think @-decorators are probably unfamiliar to many CircuitPython users. Except for `@property`, `@attr.setter` and `@staticmethod`, it appears they are only used in one of the Python guides in Adafruit_Learning_System_Guides.  I don't require you to make this change, but consider whether it's important to structure it this way and if so maybe add a comment to the effect that "putting `@timeit` at the start of a function means that when it's called, the time it takes to run will be printed in the REPL" or words to that effect.

ulab_Crunch_Numbers_Fast/benchmark.py
v923z: This was sort of stupid, because `micropython` would have worked with `time`, too. I will fix the code.

ulab_Crunch_Numbers_Fast/benchmark.py
v923z: OK.

ulab_Crunch_Numbers_Fast/benchmark.py
v923z: I can definitely change this, I simply wasn't aware of the penetration of decorators. However, I find them quite useful, so if you think that `circuitpython` users are not familiar with the topic, this could be a good tutorial subject.

v923z:
> thanks - can you run pylint on the example and update the PR :)
> https://travis-ci.com/github/adafruit/Adafruit_Learning_System_Guides/builds/161374439#L1036

Absolutely.

v923z:
@jepler I have updated the file, here are my numbers on an adafruit feather board
```
Computing the RMS value of 100 numbers
traditional          :    5.188ms [result=3535.843611]
ulab, with ndarray, some implementation in python :    0.504ms [result=3535.853624]
ulab only, with list :    0.636ms [result=3535.854340]
ulab only, with ndarray :    0.121ms [result=3535.854340]
```

Let me know if further changes are required.

jepler:
It's necessary to fix the following pylint error:
```
************* Module benchmark

ulab_Crunch_Numbers_Fast/benchmark.py 24: Trailing whitespace (trailing-whitespace)
```

after that, it can be merged.

v923z:
> It's necessary to fix the following pylint error:
> 
> ```
> ************* Module benchmark
> 
> ulab_Crunch_Numbers_Fast/benchmark.py 24: Trailing whitespace (trailing-whitespace)
> ```
> 
> after that, it can be merged.

Sorry, I forgot about pylint. It's done now.

jepler:
Looks like the CI is failing for some other reason as well.  I am going to "update branch" and see if that resolves it.  Either way, you'll need to be careful to update your local branch from github before doing further work if some is required.

v923z:
Yeah, I see that Travis is still unhappy. pylint reported the single trailing space at line 24, and I thought that that would fix it.

jepler:
Thanks for sticking with this @v923z !  And thanks for your first contribution into an Adafruit guide!

v923z:
> Thanks for sticking with this @v923z ! And thanks for your first contribution into an Adafruit guide!

It was my pleasure. It's a pity that it generated so much trouble for you. You might still have to update the text on https://learn.adafruit.com/ulab-crunch-numbers-fast-with-circuitpython/a-simple-benchmark As far as I see, that is not under version control. At least, not public.

jepler:
That's right, guide "authoring" is done via a web interface on learn.adafruit.com.  If you had an interest in authoring guides we could discuss it further on Discord.  In any case, getting those figures updated is on my list of things to do.

PR #1090, opened 2020-04-25
rounding temperature as last operation

dglaude:
To avoid loosing precision the rounding should take place after the conversion from Centigrade to Fahrenheit.
I also believe that this conversion is present in one or some CP library, so it could be reused.

WARNING: code not tested
(well I compared the math on another board)


dglaude:
Done as @dhalbert suggested.

Issue #1091, opened 2020-04-27
bug with LONNNNNNNG quotes


TonyLHansen:
The code for the FeatherWing Quote Display has a bug that shows up when using a Huzzah32. If the quote happens to more than 261 characters, you get a board fault. The test string on line 195 is about 40 characters too short to demonstrate the issue. The current set of quotes happens to have two that are longer than that test string.

ladyada:
hi please indicate the exact project URL

TonyLHansen:
my apologies. The problem is in Feather_ePaper_Quotes, found in [https://github.com/adafruit/Adafruit_Learning_System_Guides/tree/master/Feather_ePaper_Quotes](Feather_ePaper_Quotes) and as described in [https://learn.adafruit.com/epaper-display-featherwing-quote-display](epaper-display-featherwing-quote-display).

The value for the longest string appears to be less deterministic than I though. I just downloaded a fresh copy of adafruit_feather_quote, uncommented the jsonstring value on line 195, and started appending characters to the text field within it, as in "--23456789--23456789--23456789--23". At some point it will start generating faults such as  this:

`
Guru Meditation Error: Core  1 panic'ed (Unhandled debug exception)
Debug exception reason: Stack canary watchpoint triggered (loopTask) 
Core 1 register dump:
PC      : 0x401600c1  PS      : 0x00060636  A0      : 0x801601b8  A1      : 0x3ffb0070  
A2      : 0x00000000  A3      : 0x00000164  A4      : 0x00000164  A5      : 0x00000000  
A6      : 0x00000164  A7      : 0x00000000  A8      : 0x00000000  A9      : 0x3ffc1224  
A10     : 0x3ffc1224  A11     : 0x00005118  A12     : 0x00000027  A13     : 0x00000164  
A14     : 0x00000164  A15     : 0x40160164  SAR     : 0x00000010  EXCCAUSE: 0x00000001  
EXCVADDR: 0x00000000  LBEG    : 0x400014fd  LEND    : 0x4000150d  LCOUNT  : 0xffffffff  

[...]

PR #1093, opened 2020-04-27
Increased buffer in wrapWord() to prevent crash

cogliano:
This fixes issue #1091


ladyada:
nice!

TonyLHansen:
I suggest you also change the strcpy() statement a few lines later from this:

      strcpy(buff,str);

to this:

   if (strlen(str) >= sizeof(buff)-1) {  // protect against buffer overflow
      strncpy(buff, str, sizeof(buff)-2);
      buff[sizeof(buff)-1] = '\0';
    } else {
      strcpy(buff,str);
    }

That will also future proof the code.

Because of the small memory on the microcontroller, you could then drop the size of buff back down a bit so you don't have such a large buffer floating around that is empty most of the time.

PR #1095, opened 2020-04-28
first commit

jedgarpark:
Not sure if the ```global char_count, current_mode``` line is a problem in these?

### DIFF
--- a/Baudot_TTY/baudot_tty_ble.py
+++ b/Baudot_TTY/baudot_tty_ble.py
@@ -1,0 +1,230 @@
+### Baudot TTY Message Transmitter
+### Bluefruit Connect UART mode to send messages to CLUE for audio
+### tramsission to TTY machine.
+
+### The 5-bit mode is defined in ANSI TIA/EIA-825 (2000)
+### "A Frequency Shift Keyed Modem for use on the Public Switched Telephone Network"
+
+import time
+import math
+import array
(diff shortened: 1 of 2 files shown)


caternuson:
I'm OK with just using global and telling pylint to ignore.

Did you run this through black? Curious what spread out the line formatting so much.

jedgarpark:
Yes, that was from running it through Black @caternuson

Baudot_TTY/baudot_tty_ble.py
```diff
@@ -0,0 +1,230 @@
...
+    baudot_start()  # send start bit tone
+    for i in range(5):  # send each bit of the character
+        bit = (value >> i) & 0x01  # bit shift and bit mask to get value of each bit
+        baudot_bit(tones[bit])  # send each bit, either 0 or 1, of a character
+    baudot_stop()  # send stop bit
+    baudot_carrier()  # not to spec, but works better to extend carrier
+
+
+def send_message(text):
+    global char_count, current_mode  # pylint: disable=global-statement
```
jepler, line 176: I think this pylint disable is fine.  You _could_ re-organize this as a class with properties but it's not likely to be worth the effort.

jedgarpark: If you're OK I'm OK.

PR #1101, opened 2020-05-05
metronome first commit

jedgarpark:


CLUE_Metronome/clue_metronome.py
```diff
@@ -0,0 +1,209 @@
...
+# constants for sine wave generation
+SIN_LENGTH = 100  # more is less choppy
+SIN_AMPLITUDE = 2 ** 15  # 0 (min) to 32768 (max)
+SIN_OFFSET = 32767.5  # for 16bit range, (2**16 - 1) / 2
+DELTA_PI = 2 * math.pi / SIN_LENGTH  # happy little constant
+
+sine_wave = [
+    int(SIN_OFFSET + SIN_AMPLITUDE * math.sin(DELTA_PI * i)) for i in range(SIN_LENGTH)
+]
+tones = (
```
ladyada, line 31: can you use simpleio tone instead of making sine waves?

jedgarpark: Ah, I'll try that! My head was stuck back in TTY machine tone land...

jedgarpark: @ladyada this uses simpleio tone now.

PR #1107, opened 2020-05-15
Make it easier to pylint locally, put advice about it in README

jepler:


jepler:
hmmm .. this doesn't install the same version of pylint, so it sometimes gives different (additional) messages.  It might also miss messages in some circumstances.  So I'm not actually sure this is a good idea unless we upgrade pylint for the travis checks as well.

TheKitty:
@jepler - would you please review this older PR to see where things are at. Thanks.

PR #1113, opened 2020-05-18
updated BLE vibration bracelet

BlitzCityDIY:
Updated code for the BLE Vibration Bracelet Learn Guide:

-Uses adafruit_led_animation library now for NeoPixel colors
-also fixing the double ".py.py" file name


djecken:
tested code update, working nicely!

PR #1115, opened 2020-05-19
First tilemap game

FoamyGuy:
This adds the project files for the Tilemap Game Guide


FoamyGuy:
Everything is passing pylint now. I believe the guide is ready for moderation, the URL is: https://learn.adafruit.com/creating-your-first-tilemap-game-with-circuit-python/

makermelissa:
So far I've tested on the PyBadge LC and when it requests me to press the A button, I need to actually hit the B button.

makermelissa:
Button order appears to be correct on the PyGamer.

makermelissa:
It's also backwards of the Edge Badge

makermelissa:
The only other thing I noticed is that if you move Blinka out of bounds, the game crashes. I just removed a section of wall to test this. I only mention this because you are teaching people to design their own levels and they may not completely enclose Blinka like in your levels and run into this. Otherwise it seems good.

FoamyGuy:
Thank you for catching those! Both are fixed with the new commit.

makermelissa:
Thanks for making the changes. I tested them and it's working great. The only other suggestion I have is that the castle_group and sprite_group are not needed. The tilegrids can be added directly to the main group since you're not using any scaling.

FoamyGuy:
Thank you Melissa. The new commit removes the extra groups for castle and sprites. Everything is put into the main group now. 

This required one other small change: load the map before before showing the map win and lose splash screens. Previously Blinka got hidden behind the splash because the sprite was in a group behind. With Blinka sprite in the main group it got shown on top of the splash because `load_map()` started executing and removed/re-added the sprite while the splash was showing.

FoamyGuy:
Thanks Melissa! 

I will work on getting the github embed elements in the Learn Guide fixed later tonight.

PR #1121, opened 2020-05-19
fixed RTC type mismatch

andydoro:
different commands for different RTC hardware, corrected to match the Learn Guide.


andydoro:
fixed Datetime scope issue

PR #1124, opened 2020-05-24
Match latest Adafruit LED Animation library

outoftolerance:
This update fixes the code to work with the latest version of the Adafruit CircuitPython LED Animation library as the library has been split up from a single file into separate files and classes which now need to be imported differently.

**Scope**: 

- Alters the import statements at the start of the file to work with new class structure.

**Limitations**: 

- If the class structure changes again this will become out of date. 
- It is also not compatible with the older class structure.

**Testing**:

- I had to make these changes to build the Infinity Cube project which I now have fully functioning with only these changes made.


outoftolerance:
it would appear the failed test is due to a test setup issue

PR #1129, opened 2020-05-26
Update code for PyPortal IoT Plant Monitor

jimbobbennett:
This is an updated sample for using the PyPortal with Azure IoT Central. This example uses the new Azure IoT SDK that was recently published.

This PR is raised as draft and will need to go live when updates to the ALS module are ready, but not before.


.pylintrc
```diff
@@ -52,7 +52,7 @@ confidence=
...
...acking-in-except,old-raise-syntax,backtick,long-suffix,old-ne-operator,old-octal-literal,import-star-module-level,raw-checker-failed,bad-inline-option,locally-disabled,locally-enabled,file-ignored,suppressed-message,useless-suppression,deprecated-pragma,apply-builtin,basestring-builtin,buffer-builtin,cmp-builtin,coerce-builtin,execfile-builtin,file-builtin,long-builtin,raw_input-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,no-absolute-import,old-division,dict-iter-method,dict-view-method,next-method-called,metaclass-assignment,indexing-exception,raising-string,reload-builtin,oct-method,hex-method,nonzero-method,cmp-method,input-builtin,round-builtin,intern-builtin,unichr-builtin,map-builtin-not-iterating,zip-builtin-not-iterating,range-builtin-not-iterating,filter-builtin-not-iterating,using-cmp-argument,eq-without-hash,div-method,idiv-method,rdiv-method,exception-message-attribute,invalid-str-codec,sys-max-int,bad-python3-import,deprecated-string-function,deprecated-str-translate-call,import-error,missing-docstring,invalid-name,bad-whitespace,consider-using-enumerate,unexpected-keyword-arg
```
jimbobbennett, line 55: The reason for adding `unexpected-keyword-arg` is to avoid errors when explicitly using keyword args to make the code clearer. In this case using `is_celsius=True` explicitly as this code is called out in the guide.

PyPortal_Azure_Plant_Monitor/azure_gfx_helper.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, is_celsius=True):
         self.temp_label.y = 65
         self._text_group.append(self.temp_label)
 
-        self.temp_text = Label(self.data_font, max_glyphs=10)
+        self.temp_text = Label(self.data_font, max_glyphs=10, text="")
```
jimbobbennett, line 59: This is a workaround for https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/45. This can be reverted once the fix is released.

PyPortal_Azure_Plant_Monitor/azure_gfx_helper.py
```diff
@@ -66,19 +66,19 @@ def __init__(self, is_celsius=True):
...
         self._text_group.append(self.moisture_label)
 
-        self.moisture_text = Label(self.data_font, max_glyphs=10)
+        self.moisture_text = Label(self.data_font, max_glyphs=10, text="")
         self.moisture_text.x = 200
         self.moisture_text.y = 175
         self._text_group.append(self.moisture_text)
 
-        self.azure_status_text = Label(self.main_font, max_glyphs=15)
+        self.azure_status_text = Label(self.main_font, max_glyphs=15, text="")
```
jimbobbennett, line 74: This is a workaround for https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/45. This can be reverted once the fix is released.

PyPortal_Azure_Plant_Monitor/azure_gfx_helper.py
```diff
@@ -66,19 +66,19 @@ def __init__(self, is_celsius=True):
         self.moisture_label.y = 135
         self._text_group.append(self.moisture_label)
 
-        self.moisture_text = Label(self.data_font, max_glyphs=10)
+        self.moisture_text = Label(self.data_font, max_glyphs=10, text="")
```
jimbobbennett, line 69: This is a workaround for https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/45. This can be reverted once the fix is released.

ladyada:
@brentru wanna review? :)

PyPortal_Azure_Plant_Monitor/azure_gfx_helper.py
brentru: @jimbobbennett  This issue was fixed here: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/45, could you please update this PR to reflect that?

PyPortal_Azure_Plant_Monitor/code.py
```diff
@@ -31,18 +38,36 @@
 esp32_reset = DigitalInOut(board.ESP_RESET)
 spi = busio.SPI(board.SCK, board.MOSI, board.MISO)
 esp = adafruit_esp32spi.ESP_SPIcontrol(spi, esp32_cs, esp32_ready, esp32_reset)
+
+# Set up the WiFi manager with a status light to show the WiFi connection status
 status_light = neopixel.NeoPixel(board.NEOPIXEL, 1, brightness=0.2)
 wifi = adafruit_esp32spi_wifimanager.ESPSPI_WiFiManager(esp, secrets, status_light)
+wifi.connect()
```
brentru, line 45: Could you add a `print` before this to show the user that the WiFi is connecting, and a `print` statement after displaying it's connected? It's often hard to determine the state of the code when the display is off.

PyPortal_Azure_Plant_Monitor/code.py
```diff
@@ -31,18 +38,36 @@
...
+
+# Time setup, needed to authenticate with Azure IoT Central
+ntp = NTP(esp)
+while not ntp.valid_time:
```
brentru, line 49: may want to print `Could not get valid time, retrying in 5s...`

brentru: something like: https://github.com/adafruit/Adafruit_CircuitPython_NTP/blob/master/examples/ntp_simpletest.py#L34

PyPortal_Azure_Plant_Monitor/code.py
```diff
@@ -52,15 +77,27 @@
...
+        }
+        device.send_telemetry(json.dumps(message))
+        device.loop()
+
         gfx.display_azure_status('Data sent!')
         print('Data sent!')
     except (ValueError, RuntimeError) as e:
         print("Failed to get data, retrying\n", e)
         wifi.reset()
+        wifi.connect()
```
brentru, line 98: Q: Does `wifi.reset()` not handle reconnections? We may want to add a `reconnect` method to ESP32SPI's `wifimanager` if it doesn't

jimbobbennett: It doesn't reconnect automatically - I tested this out to make sure.

Raised https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/issues/100 to add this, will try to implement it later today if I get a chance.

jimbobbennett:
@brentru - updated with your requested changes, and updated the draft pages in the ALS guide to match.

jimbobbennett:
Thanks @brentru - to avoid confusion I assume we can hold off merging till the ALS guide changes are approved?

brentru:
@jimbobbennett  Sure, I'll hold off until the guide is updated.

PyPortal_Azure_Plant_Monitor/code.py
jimbobbennett: @brentru - adding to this. The WiFi manager doesn't reconnect, but all its methods will connect if it's not connected, meaning that after a `reset`, a call to `get` will work.

When using the socket like I am in the Azure library, the connection is not remade.

Add more comments on this to the issue mentioned above. Would be good to get your thoughts.

PR #1139, opened 2020-06-09
updating pyportal guitar tuner

BlitzCityDIY:
Updating the PyPortal Guitar Tuner code to use default_bg with the PyPortal rather than create a tilegrid.


djecken:
tested code update, working good! running on Adafruit CircuitPython 5.3.0 on 2020-04-29; Adafruit PyPortal with samd51j20

PR #1141, opened 2020-06-12
Circuit Playground auto bike brake light

dherrada:
Works best with CPB because it's faster, but it still works fine on the CPX

### DIFF
--- a/Circuit_Playground_Brake_Light/code.py
+++ b/Circuit_Playground_Brake_Light/code.py
@@ -37,7 +37,7 @@ while True:
                 cp.pixels.brightness = 0.25
                 start = time.monotonic()
                 sleep = True
-        i +=1
+        i += 1
 
     # sleep variable is for short circuiting
     elif not sleep or time.monotonic() - start > 0.4:
--- a/Circuit_Playground_Brake_Light/code.py
+++ b/Circuit_Playground_Brake_Light/code.py
@@ -3,14 +3,20 @@ import math
 from adafruit_circuitplayground import cp
 
 brightness = 0
-l10 = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
-l50 = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0,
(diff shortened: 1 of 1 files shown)


Circuit_Playground_Brake_Light/code.py
```diff
@@ -0,0 +1,53 @@
...
+            # Detects shake. Due to the very low shake threshold, this alone would have very low
+            # specificity. This was mitigated by having it only run when the acceleration
+            # difference is greater than 1 m/s^2 at least 3 times in a row.
+            if not cp.shake(shake_threshold=10):
+                cp.pixels.brightness = 1
+                start = time.monotonic()
+                sleep = True
+        i += 1
+
+    # sleep variable is for short circuiting. Not really necessary, but it makes things run faster
```
kattni, line 44: Using `sleep` as the variable name could be confused with `time.sleep()`.

Circuit_Playground_Brake_Light/code.py
```diff
@@ -0,0 +1,53 @@
+import time
+import math
+from adafruit_circuitplayground import cp
+
+brightness = 0
+l10 = [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```
siddacious, line 6: I'd change these to more meaningful names and have a comment about their use

Circuit_Playground_Brake_Light/code.py
```diff
@@ -0,0 +1,53 @@
...
+
+i = 0
+cp.pixels.fill((255, 0, 0))
+
+light_on = False
+
+while True:
+    x, y, z = cp.acceleration
+
+    # moving average n=10, not super smooth, but it substantially lowers the amount of noise
```
siddacious, line 21: The rest of the code could use more explanation of what is going on and why. If you are going to go though each section of the code in the guide,  this is ok as is (though comments with more explanation never hurt)

dherrada: My plan was to explain it in the guide, but I'll add some more explanations

PR #1145, opened 2020-06-16
Added zip of STLs

dherrada:
Just adding my 3d printer files so people can download them easily.


ladyada:
please commit them unzipped, so they can be rendered in github

dherrada:
@ladyada Ok, will do

dherrada:
Using thingiverse instead

PR #1149, opened 2020-06-17
Adding combo board ino example

kattni:


ladyada:
@kattni plz add `.uno.test.only` file so it doesnt test all the platforms :)

PR #1157, opened 2020-06-29
remove LSM303

ladyada:


ladyada:
@kattni the new arduino ide doesn't like it when the name of the arduino sketch doesnt match the folder so i renamed all the bad files. they'll hvae to be `re embedded` in the matching guides.  but! CI passes now. so its a fair trade :)
please review and update the guide file embeds once merged, sound like a plan?

kattni:
@ladyada All guides have been updated.

ladyada:
huzzah thank u

PR #1158, opened 2020-06-29
Add Arduino example for Feather Sense

caternuson:
Basically an Arduino version of the CircuitPython example.

CI is trying to install a deprecated library (LSM303) which is causing it to fail?

Screenshot from 2020-06-29 15-40-17


ladyada:
please add a `.cpb.test.only` so only bluefruit is tested

PR #1163, opened 2020-07-04
Adding PyPortal Electioncal guide code

fede2cr:
This change add the directory folder for the code related to a new guide about using PyPortal to show Electioncal US dates.

It has been tested with a few random counties to make sure it doesn't crashes or looks funny.

If you want to test, while the guide is being write you can just copy the files over to your pyportal, add the missing libraries from the bundle if needed, and create a secrets.py file as with other PyPortal projects.

Rename electioncal.py to code.py, and then just change your state/county in the top of this file.

Thanks in advance


fede2cr:
Thanks for the software test @FoamyGuy.
This will be added to the Troubleshooting page as you suggestion.
In this case the request of the json file is handled by the PyPortal library, and jsons cannot be added as replies as the website is static pages on github, so the reply is html.
You are right, it is easy yo catch it, I'm adding the code right now.

fede2cr:
With the new code, if there is an error with state or county, it now prints a nice error message: (my cpy is in spanish, and mu-console has trouble with utf)
Retrieving data...Reply is OK!
Error loading JSON data: Please check the configuration of county and state, in code.py
Traceback (ultima llamada reciente):
  Archivo "code.py", lÃ­nea 53, en <module>
  Archivo "code.py", lÃ­nea 50, en <module>
  Archivo "electioncal_graphics.py", lÃ­nea 76, en load_data
  Archivo "electioncal_graphics.py", lÃ­nea 72, en load_data
ValueError: error de sintaxis en JSON



Presiona cualquier tecla para entrar al REPL. Usa CTRL-D para recargar.

PR #1169, opened 2020-07-13
Update waterfall.py

CedarGroveStudios:
ulab function doesn't support indexing and throws an error in CircuitPython 5.3.0.


CedarGroveStudios:
resubmitted with error information

PR #1177, opened 2020-07-15
Create circuit_playground_bluefruit_sensornet.py

dglaude:
This could be an addition to this guide: https://learn.adafruit.com/bluetooth-le-broadcastnet-sensor-node-raspberry-pi-wifi-bridge

It is "untested" (the code works on the CPB, but I have not configured my Pi yet).

One limitation is that the circuitplayground library has a sound_level propertie, but I did not find a matching propertie in the broadcastnet library.

### DIFF
--- a/BLE_Sensornet/circuit_playground_bluefruit_sensornet.py
+++ b/BLE_Sensornet/circuit_playground_bluefruit_sensornet.py
@@ -1,0 +1,17 @@
+"""This uses the Circuit Playground Bluefruit as a Bluetooth LE sensor node."""
+
+import time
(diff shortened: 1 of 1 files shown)


dglaude:
Maybe this is a duplicate and is not needed if the guide does not have a section on CPB.

I found similar code in the broadcastnet library example: https://github.com/adafruit/Adafruit_CircuitPython_BLE_BroadcastNet/blob/master/examples/ble_broadcastnet_cpb.py

tannewt:
@jedgarpark Do you want to add this an example to the guide? Or is the library example enough?

PR #1180, opened 2020-07-19
Make the Twitter_API example work

dglaude:
(1) on the import section, I prefer to use adafruit_binascii library than a version specific to this example
(2) apparently requests is outside of esp32spi
(3) the code to compute or ask Twitter the bearer_token fail for me ... but when I created my Twitter Developper account, in addition to the 'twitter_api_key' and 'twitter_secret_key', there was a bearer token that I have put in my secrets.py file :

'twitter_bearer_token' : 'verylongstring'

The guide should be updated to take that into account I guess.

I have not deleted the function get_bearer_token(), but it is not in use anymore.

Here is the error I get (I did not try to fix, but I had it multiple time).

  File "main.py", line 1, in <module>
  File "main_twitter.py", line 77, in <module>
  File "main_twitter.py", line 47, in get_bearer_token
  File "adafruit_requests.py", line 296, in post
  File "adafruit_requests.py", line 251, in request
  File "adafruit_requests.py", line 235, in request
  File "adafruit_esp32spi/adafruit_esp32spi_socket.py", line 102, in readline
RuntimeError: Didn't receive full response, failing out


dastels:
What version of CircuitPython and the libraries are you using?

dglaude:
I am on 6.0.0-alpha1 for the firmware and 20200718 for the library.

So that would be the following library that are relevant:
* https://github.com/adafruit/Adafruit_CircuitPython_Requests/releases/tag/1.4.3
* https://github.com/adafruit/Adafruit_CircuitPython_ESP32SPI/releases/tag/3.4.2
* https://github.com/adafruit/Adafruit_CircuitPython_binascii/releases/tag/1.2.2

I am new to Twitter API, and it is not sure that I know what I am doing. So take my remarks with some distance, I just feel things may have change both on Twitter side and on CircuitPython library and core. So I had to make minor change to make it work for me.

dastels:
I still see adafruit_requests.mpy as being directly in the lib directory.

I also don't see a change in the requirement for the bearer token fetch.

I am getting the same 
```
File "adafruit_esp32spi/adafruit_esp32spi_socket.py", line 102, in readline
RuntimeError: Didn't receive full response, failing out
```

dglaude:
Hi @dastels , I downgraded my PyPortal to 5.3.1 (keeping the same library version) and I have that same error. So there must be something going wrong with esp32spi. Since this is currently in rewriting for 6.0, maybe I should not worry yet.

The story about the bearer token fetch is that I did not have to fetch it using the code here... I got one such key directly from the same Twitter interface were I get the other keys. But sure, in Twitter documentation I found out the same process to build that bearer token.

jepler (commented):
Looks fine, no testing performed.

jepler (approved):

Twitter_API/code.py
```diff
@@ -14,11 +14,11 @@
 
 #pylint:disable=invalid-name
 import time
-import binascii
+import adafruit_binascii as binascii
```
brentru, line 17: @dglaude Since we're using `adafruit_binascii`, could you please create a commit removing the old `binascii` file in this directory as to not confuse users: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/933c73cdefd097dbadc58ace9a865272d2a21298/Twitter_API/binascii.py

Twitter_API/code.py
```diff
@@ -71,7 +71,7 @@ def get_bearer_token():
                     caption_position=(5, 210),
                     caption_color=0x808080)
 
-bearer_token = get_bearer_token()
+bearer_token = secrets['twitter_bearer_token']
```
brentru, line 74: @dglaude  This works and we don't need to generate the token anymore - hurray! Could you please remove the `get_bearer_token` and `halt_and_catch_fire` functions since they are unnecessary now and not called elsewhere?

brentru (changes requested):
@dglaude  I just tested this on a PyPortal (`Adafruit CircuitPython 6.1.0-beta.2 on 2020-12-03; Adafruit PyPortal with samd51j20`) and a fresh twitter developer account.

I have some feedback to clean this up before I can approve and merge. I'll update the guide with new screenshots, too.

PR #1185, opened 2020-07-22
New Guide - rock, paper, scissors game for CLUE and CPB.

kevinjwalters:
Some CircuitPython files for a new guide. Four directories, three different versions of the game plus unit tests:

- `very-simple`
- `simple`
- `advanced` - this includes an `rps` subdirectory holding `wav` files and a `bmp` file
- `tests` - two units tests which can run under CPython


kevinjwalters:
**_Please don't merge yet - I've got a little bit more minor tidying to do, will do this today._**

kevinjwalters:
Good to go, now.

kevinjwalters:
@TheKitty might be interested in this.

TheKitty:
@dhalbert - would you please double check this use of Bluetooth between devices per this guide in moderation https://learn.adafruit.com/admin/guides/2998/editor

kevinjwalters:
FYI, there's some notes on the innards in the draft Learn Guide with current title of "CLUE Rock, Paper, Scissors Game using Bluetooth"

kevinjwalters:
I've just improved the error reporting in case the user has missing audio files. Will have a commit for that on the way within 20 mins.

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
```diff
@@ -0,0 +1,556 @@
...
+
+    # Inputs
+    # buttons reversed if it is used upside-down with Gizmo
+    _button_a = digitalio.DigitalInOut(board.BUTTON_A)
+    _button_a.switch_to_input(pull=digitalio.Pull.DOWN)
+    _button_b = digitalio.DigitalInOut(board.BUTTON_B)
+    _button_b.switch_to_input(pull=digitalio.Pull.DOWN)
+    if display is None:
+        button_left = lambda: _button_a.value
+        button_right = lambda: _button_b.value
```
dhalbert, line 141: Why lambdas instead of single-line def's?

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
```diff
@@ -0,0 +1,556 @@
...
+    # Return with void at True if any input is None
+    try:
+        mine_lc = mine.lower()
+        yours_lc = yours.lower()
+    except AttributeError:
+        return (False, False, True)
+
+    r_win = r_draw = r_void = False
+    # pylint: disable=too-many-boolean-expressions
+    if (mine_lc == "rock" and yours_lc == "rock"
```
dhalbert, line 259: Could use `in` here to check that the values are valid. Then just check `mine_lc == yours_lc`  for the draw case.

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
```diff
@@ -0,0 +1,556 @@
...
+TOTAL_ROUNDS = 3
+
+CRYPTO_ALGO = "chacha20"
+KEY_SIZE = 8  # in bytes
+KEY_ENLARGE = 256 // KEY_SIZE // 8
+
+
+def evaluateRound(mine, yours):
+    """Determine who won the game based on the two strings mine and yours.
+       Returns three booleans (win, draw, void)."""
```
dhalbert, line 249: It appears to me that this tuple always returns a tuple that's always one `True` and two `False's`. Why not just return `True`, `False`, or `None`, or one of three strings, or a enum-like value?

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
```diff
@@ -0,0 +1,237 @@
...
+    _DATA_FMT_ENC_DATA = "8s"
+
+    prefix = struct.pack(
+        _PREFIX_FMT,
+        struct.calcsize(_PREFIX_FMT) - 1,
+        MANUFACTURING_DATA_ADT,
+        ADAFRUIT_COMPANY_ID,
+        struct.calcsize("<H" + _DATA_FMT_ENC_DATA),
+        RPS_ENC_DATA_ID
+    )
```
dhalbert, line 76: You can now use `match_prefixes = (prefix1, prefix2, ...)`. You do not have to specify prefix lengths. The underlying code mashes the prefixes together, and adds the length headers.

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
```diff
@@ -0,0 +1,237 @@
...
+    # The second struct.calcsize needs to include the _DATA_FMT for some
+    # reason I either don't know or can't remember
+    prefix = struct.pack(
+        _PREFIX_FMT,
+        struct.calcsize(_PREFIX_FMT) - 1,
+        MANUFACTURING_DATA_ADT,
+        ADAFRUIT_COMPANY_ID,
+        struct.calcsize("<H" + _DATA_FMT_KEY_DATA),
+        RPS_KEY_DATA_ID
+    )
```
dhalbert, line 128: Can use `match_prefixes`.

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
```diff
@@ -0,0 +1,237 @@
...
+    _PREFIX_FMT = "<B" "BHBH"
+
+    prefix = struct.pack(
+        _PREFIX_FMT,
+        struct.calcsize(_PREFIX_FMT) - 1,
+        MANUFACTURING_DATA_ADT,
+        ADAFRUIT_COMPANY_ID,
+        struct.calcsize("<H" + _DATA_FMT_ROUND),
+        RPS_ROUND_ID
+    )
```
dhalbert, line 176: ditto

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
```diff
@@ -0,0 +1,237 @@
...
+    prefix = struct.pack(
+        _PREFIX_FMT,
+        struct.calcsize(_PREFIX_FMT) - 1,
+        MANUFACTURING_DATA_ADT,
+        ADAFRUIT_COMPANY_ID,
+        ##struct.calcsize("<H" + _SEQ_FMT + _DATA_FMT),
+        struct.calcsize("<H" + _DATA_FMT),
+        ##struct.calcsize("<H"),
+        GM_JOIN_ID
+    )
```
dhalbert, line 222: ditto

CLUE_Rock_Paper_Scissors/advanced/rps_comms.py
```diff
@@ -0,0 +1,361 @@
...
+    # And make a list of sequence numbers already acknowledged
+    acks = {}
+    for addr_text, adsnb_per_addr in received_ads_by_addr.items():
+        if cls_send_ad in [type(andb[0]) for andb in adsnb_per_addr]:
+            send_ad_rxs[addr_text] = True
+
+        # Pick out any Advertisements with an ack field with a value
+        acks_thisaddr = list(filter(lambda adnb: hasattr(adnb[0], "ack")
+                                    and isinstance(adnb[0].ack, int),
+                                    adsnb_per_addr))
```
dhalbert, line 293: I believe this could be a list comprehension with an `if`; might be a little easier to read.

CLUE_Rock_Paper_Scissors/advanced/rps_crypto.py
```diff
@@ -0,0 +1,100 @@
...
+def strUnpad(text_as_bytes, pad=0):
+    """Convert a bytes or bytearray
+       to a str removing trailing characters matching pad."""
+    if pad is not None:
+        end_ex = len(text_as_bytes)
+        while end_ex > 0 and text_as_bytes[end_ex - 1] == pad:
+            end_ex -= 1
+
+    # bytearray does not have decode method so need to convert here
+    return bytes(text_as_bytes[0:end_ex]).decode("utf-8")
```
dhalbert, line 49: Could you use `rstrip()` here?

kevinjwalters: Yes, this originated from `bytearray()` not having much to offer, all the cool functionality is in `bytes()`. But, I then needed `decode` too so I do end up going via `bytes()` and need to revisit that, thanks.

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
kevinjwalters: I think it's `win`, `draw`, `lose`, `something-is-wrong`. I'm certainly up for rewriting that as I can contrast it with the the other similar one in code discussion. I think that return format just matched up with the counters in the main loop. What's CircuitPython's alternative to `from enum import Enum` ?

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
kevinjwalters: No particular reason. The **pylint** police will be after me if I literally do a one-line `def`. Should I change these to

```
def button_left():
    return _button_a.value
```

then?

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
dhalbert: You could just use integers as an enum, either in a class a class or variables, or globals.

CLUE_Rock_Paper_Scissors/advanced/rps_crypto.py
kevinjwalters: Done

CLUE_Rock_Paper_Scissors/advanced/rps_comms.py
kevinjwalters: Done

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
kevinjwalters: Done

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
kevinjwalters: `evaluateRound` is now smaller and returns globals which are `const()` integers in the main program. I've left the booleans passed to the `RPSDisplay` method as is...

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
kevinjwalters: I have migrated to match_prefixes.

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
kevinjwalters: I have migrated to match_prefixes.

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
kevinjwalters: I have migrated to match_prefixes.

CLUE_Rock_Paper_Scissors/advanced/rps_advertisements.py
kevinjwalters: I have migrated to match_prefixes.

Also did the one in the simple version.

kevinjwalters:
- [ ] I picked up `20200806` libs and I now have one text Label that has changed location and is half off the screen. I need to sort that out.

kevinjwalters:
On the format of the application's `Advertisement` various packets I'm not sure. I carefully selected ID values to achieve a certain order based on currrent hash table algo. That's clearly fragile, bad practice and almost guaranteed to break if/when CircuitPython is changed to match the CPython `dict()` ordering. I think I could use the multi-field feature in `ManufacturerDataField` but that wouldn't be elegant for optional fields.

I intended to watch this space and update the code with a better solution when one appeared or I found one.

If ordering _within_ `ManufacturerData` could be controlled then my use of it would be ok. There's a few ways to offer that if that's generally desirable:

1. Order as field values are set.
2. Order as fields are declared in the sub-class of `Advertisement`
3. Some explicit specification of full / partial ordering requirement

I like the sound of **2** myself but **1** happens to be very easy to implement as I think it could be achieve with just replaced `{}` with an `OrderedDict()` at https://github.com/adafruit/Adafruit_CircuitPython_BLE/blob/9efa23a823df695421dcc766832f397a236ac1f0/adafruit_ble/advertising/standard.py#L223 ?

TheKitty:
@kevinjwalters Is it at the merge point now?

kevinjwalters:
@TheKitty Almost, would be good to get a view on https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1185#issuecomment-670944106 from @dhalbert

CLUE_Rock_Paper_Scissors/advanced/clue-multi-rpsgame.py
kevinjwalters: Do you want to me to change these to `def`s?

dhalbert: If you want to explain the code, it's probably easier not to have to explain lambdas.

kevinjwalters: There's a few more. I use them for adapting the args on some callback functions too further down. And one lurks in `rps_display.py` for `key` in `sort()` use for NeoPixel displaying.

Quick hacky survey shows Learn code does use these although 4 of those including this one are mine.

```
$ for fred in *; do echo ${fred} $(cat $(find ${fred} -type f -name '*.py' -print) /dev/null | fgrep -c lambda:); done | grep -v ' 0$'
CircuitPython_JEplayer_mp3 4
CLUE_Dice_Roller 2
CLUE_Metal_Detector 5
CLUE_Rock_Paper_Scissors 18
CLUE_Sensor_Plotter 5
CPB_Quick_Draw_Duo 2
Disc_Step_Sequencer 3
TFT_Gizmo_Turtle_Graphics 7
ulab_Crunch_Numbers_Fast 4
```

I have a slight preference to leave as is but I'm ok with stripping them out too.

dhalbert: Lambdas are useful when you don't want to bother to name a function. In this case, it's just a way of circumventing pylint complaints, and I'd prefer using just `def`'s. The black formatter adds a lot of other extra lines too. We run all library code through black; you might consider the same thing, though you might not like the result.

dhalbert: Right, I just marked the first instance of these; there are others.

dhalbert:
> If ordering _within_ `ManufacturerData` could be controlled then my use of it would be ok. There's a few ways to offer that if that's generally desirable:
> 
> 1. Order as field values are set.
> 2. Order as fields are declared in the sub-class of `Advertisement`
> 3. Some explicit specification of full / partial ordering requirement
> 
> I like the sound of **2** myself but **1** happens to be very easy to implement as I think it could be achieve with just replaced `{}` with an `OrderedDict()` at https://github.com/adafruit/Adafruit_CircuitPython_BLE/blob/9efa23a823df695421dcc766832f397a236ac1f0/adafruit_ble/advertising/standard.py#L223 ?

I think 1 is fine with an `OrderedDict`. But I would say three things:
1. You don't have to filter based on the Manufacturer Data fields. You can just filter on the manufacturer (or not at all), and then filter "by hand" later.
2. You don't have to adhere to that format of advertisement.
3. You can merge the Mfr Data fields into a single structure field.

kevinjwalters:
I've updated my code and thrown in some TDD using new `tests/test_rps_advertisements.py`. I put in https://github.com/adafruit/Adafruit_CircuitPython_BLE/pull/97 for the OrderedDict enhancement which passes on that test.

kevinjwalters:
@TheKitty This is good to merge now.

PR #1190, opened 2020-07-27
pixel chase game code

BlitzCityDIY:
Adding the code for the Pixel Chase Game Learn Guide.

Using 61 NeoPixels in a circle, a "chaser" pixel flies around. A "target" pixel is setup randomly in the circle and using a push button, you try and hit the target with the chaser. The colors change and speed increases with each level up. If you press the button without being on target, the game resets. Once you win all the levels, the NeoPixels go into rainbow_cycle for a bit and then a new game begins.

### DIFF
--- a/Pixel_Chase_Game/code.py
+++ b/Pixel_Chase_Game/code.py
@@ -1,0 +1,190 @@
+import time
+import random
+import board
+import neopixel
+import digitalio
+import adafruit_led_animation.color as color
+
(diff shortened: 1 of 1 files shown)


dglaude:
Not my take and I know the code is already merged...

But could the rainbow_cycle(wait) be replaced by something from Adafruit LED Animation.
That would remove the need for wheel too.

PR #1192, opened 2020-07-29
Clue sensor plotter memory fix

FoamyGuy:
This tweak to the CLUE sensor plot project will prevent the info label from using a bitmap for it's background color, but still retains the blue background by manipulating the labels palette. 

This seems to allow it to save enough memory to get back to working when used with the newest display_text release that brought in the optimization for `None` backgrounds.

This issue was discussed here: https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/61


caternuson:
Looks good to me. I can also confirm it works.
clue_test

PR #1196, opened 2020-07-31
Added astrophotography tracker code

dherrada:
The only thing that might change on this is the second time.sleep. If that changes, then step_time.py will also probably change since that would mean I got my calculations wrong. So essentially, it'd be great if someone could review it but not merge it.


dherrada:
@brentru So `step_time.py` is just a file that helps you find what delay to use. That being said, I think the way you suggested makes more sense.

dherrada:
Also, I'm much more confident that my calculations are correct for this now. Initially, I wasn't getting the speed I thought I should be getting, but it turns out I just had the microstepping pins wired incorrectly. Friday night, I ran it for exactly 6 hours, and it rotated exactly 90 degrees, which is the speed I was shooting for.

brentru:
@dherrada  The original comment mentioned reviewing but not merging, do you still want me to do this, or do you want it merged now?

dherrada:
@brentru yeah, you can merge it now. Only reason I didn't want it merged earlier is that I hadn't done a test at the actual speed it'd be running at. Thanks!

brentru:
Thanks, will merge when passing.

PR #1201, opened 2020-08-04
CLUE altimeter update

caternuson:
Slight oversight on my part. Didn't think about dealing with initial condition for NVM. Simple fix.

Re this thread:
https://forums.adafruit.com/viewtopic.php?f=65&t=167847


kevinjwalters:
For comparison, the micro:bit uses a 32bit "unique" identifier to indicate that the compass calibration data is present: https://github.com/bbcmicrobit/micropython/blob/0185d0bf99ded455b64929f2489a76bb21bf406f/source/microbit/microbitcompass.cpp#L36-L52

PR #1205, opened 2020-08-13
Catch Ctrl-C and unlock I2C bus before exiting

hjw:
Put while true loop inside a try block in order to catch ctrl-c and unlock the bus before exiting... prevents i2c bus being left in locked state. Also added comment on how to unlock the bus from the REPL.


CircuitPython_Essentials/CircuitPython_I2C_Scan.py
```diff
@@ -1,15 +1,24 @@
...
 
 while not i2c.try_lock():
     pass
 
-while True:
-    print("I2C addresses found:", [hex(device_address)
-                                   for device_address in i2c.scan()])
-    time.sleep(2)
+try:
+
```
kattni, line 17: Please delete this blank line.

CircuitPython_Essentials/CircuitPython_I2C_Scan.py
```diff
@@ -1,15 +1,24 @@
...
+    while True:
+        print("I2C addresses found:", [hex(device_address)
+            for device_address in i2c.scan()])
```
kattni, line 20: This line needs two more spaces in it to line up with the rest of the print statement.

The other option is that the print statement fits on one line. If you'd rather, you can put this on the same line as `print("I2C addresses found:"....`.

CircuitPython_Essentials/CircuitPython_I2C_Scan.py
```diff
@@ -1,15 +1,24 @@
 # CircuitPython demo - I2C scan
+# 
+# If you run this and it seems to hang, try manually unlocking
+# your I2C bus from the REPL with
+# ` >>> import board `
+# ` >>> board.I2C().unlock() `
```
kattni, line 6: I do not believe the backticks are necessary here as it is already in a comment. If you included them for a reason, please let me know what it was.

hjw: done, thought they might be necessary for auto gen docs to show as code, but in reality this code isn't used that way.

CircuitPython_Essentials/CircuitPython_I2C_Scan.py
hjw: Leaving it on two lines, added the spaces, that's the way it was originally.

ladyada:
merged, thanks @hjw !

PR #1207, opened 2020-08-14
Add hurricane tracker code

caternuson:
Here it is! Pull request like a hurricane.


PyPortal_Hurricane_Tracker/hurricane_tracker.py
```diff
@@ -0,0 +1,171 @@
...
+    # clear out existing icons
+    while len(storm_icons):
+        _ = storm_icons.pop()
+
+    # get latest storm data
+    try:
+        storm_data = pyportal.fetch()
+    except RuntimeError:
+        return
+    print("Number of storms:", len(storm_data))
```
jerryneedell, line 73: This prints # of storms but later only storms in area are displayed

PyPortal_Hurricane_Tracker/hurricane_tracker.py
```diff
@@ -0,0 +1,171 @@
...
+        if len(storm_icons) >= MAX_STORMS:
+            continue
+        # get lat/lon
+        lat = storm["latitudeNumeric"]
+        lon = storm["longitudeNumeric"]
+        # check if on map
+        if (
+            not LAT_RANGE[0] >= lat >= LAT_RANGE[1]
+            or not LON_RANGE[0] <= lon <= LON_RANGE[1]
+        ):
```
jerryneedell, line 87: Do you want to print something if storms out of area. I was confused when it said there were two storms found, but only printed information on one.

PyPortal_Hurricane_Tracker/hurricane_tracker.py
caternuson: In general, all the serial prints are debug info. For this one, was helpful to know actual storm count in raw JSON.

PyPortal_Hurricane_Tracker/hurricane_tracker.py
caternuson: Same here. If no storms end up being within the map area, then there's another debug print that should show up.

caternuson:
Awesome, thanks for testing it. Good to know it works :)

PyPortal_Hurricane_Tracker/hurricane_tracker.py
jerryneedell: This is fine.. and useful.

PyPortal_Hurricane_Tracker/hurricane_tracker.py
jerryneedell: As I said, no big deal -- it just took me a few minutes to figure out why it said there were 2 storms and only showed 1.
I saw that if there are none in range it prints a message.
I'm fine with it.

jerryneedell:
Looks like the map is only for Atlantic  -- Is that going to be specified in the Guide -- or will it be expanded for other regions as well.

caternuson:
Yep. Just Atlantic for now. Maybe add other regions later. But also - that would make for an excellent user update. Use the provided Atlantic as an example and extent from there.

The NOAA JSON source covers Atlantic, East Pacific, and Central Pacific. So those regions could be added fairly easily.

The other missing bit is a map for a Titano. Simple resize, but didn't have the hardware to test.

jerryneedell:
> Yep. Just Atlantic for now. Maybe add other regions later. But also - that would make for an excellent user update. Use the provided Atlantic as an example and extend from there.
> 
> The NOAA JSON source covers Atlantic, East Pacific, and Central Pacific. So those regions could be added fairly easily.
> https://nws.weather.gov/products/PDD/PDD_ExpTropicalCycloneJSONFile_2019.pdf
> 
> The other missing bit is a map for a Titano. Simple resize, but didn't have the hardware to test.

ooh -- I have a Titano -- I suppose I should learn how to do the resize...

caternuson:
Oh, they're different aspect ratios:

- Pynt / Original: 320x240 = 4:3
- Titano : 480x320 = 3:2

so actually a little more than just a resize.

PR #1211, opened 2020-08-17
update to Daily Cheer Automaton to work with updated version of DST library

andydoro:
update to Daily Cheer Automaton arduino version to work with updated version of DST library, must define US or EU DST rules.


ladyada:
merged, thanx!

PR #1218, opened 2020-08-22
Update single_matrix.py

dglaude:
Blind change to fix change in adafruit_lsm6ds.
Totally untested, just based on pattern matching.


ladyada:
thanks!

PR #1221, opened 2020-08-22
Update powerglove_ble_midi.py

dglaude:
Blind change totally untested only based on pattern matching to fix the change in adafruit_lsm6ds.


jedgarpark:
Thanks @dglaude I've fixed it, also renamed the file to code.py so people who download the Project Bundle don't need to rename it.

PR #1228, opened 2020-08-26
https://learn.adafruit.com/sword-and-wand-prop-effects-with-circuit-playground

ladyada:


ladyada:
@jedgarpark plz approve and then update the guide to use this instead of a zip

PR #1229, opened 2020-08-28
first commit

jedgarpark:
main code and graphics class code plus fonts, background, and icons.


Weather_Display_Matrix/openweather_graphics.py
```diff
@@ -0,0 +1,205 @@
...
+            row = None
+            for index, icon in enumerate(icon_map):
+                if icon == icon_name[0:2]:
+                    row = index
+                    break
+            column = 0
+            if icon_name[2] == "n":
+                column = 1
+            if row is not None:
+                print(column, row)
```
makermelissa, line 170: Maybe you can remove this print statement. I accidentally left it in for debugging and forgot to remove it.

Weather_Display_Matrix/weather_display_matrix.py
```diff
@@ -0,0 +1,88 @@
...
+matrix = Matrix()
+network = Network(status_neopixel=board.NEOPIXEL, debug=True)
+if UNITS == "imperial":
+    gfx = openweather_graphics.OpenWeather_Graphics(
+        matrix.display, am_pm=True, units="imperial"
+    )
+elif UNITS == "metric":
+    gfx = openweather_graphics.OpenWeather_Graphics(
+        matrix.display, am_pm=True, units="metric"
+    )
```
makermelissa, line 59: This could be reduced to:
```
if UNITS == "imperial" or UNITS == "metric":
    gfx = openweather_graphics.OpenWeather_Graphics(
        matrix.display, am_pm=True, units=UNITS
    )

File: Weather_Display_Matrix/openweather_graphics.py [hunk 3]
```diff
@@ -0,0 +1,205 @@
...
+            units="imperial"
+    ):
+        super().__init__(max_size=3)
+        self.am_pm = am_pm
+        if units == "metric":
+            self.celsius = True
+            self.meters_speed = True
+        else:
+            self.celsius = False
+            self.meters_speed = False
```
makermelissa, line 42: Would it make more sense to have units as a boolean like `metric_units=False` and then you could set:
```
self.celsius = metric_units
self.meters_speed = metric_units
```

Weather_Display_Matrix/openweather_graphics.py
jedgarpark: removed!

Weather_Display_Matrix/weather_display_matrix.py
jedgarpark: Done!

Weather_Display_Matrix/openweather_graphics.py
jedgarpark: we had discussed yesterday how keeping it the current way is descriptive and allows for someone to add the default Kelvin mode if they want.

PR #1242, opened 2020-09-20
PyPortal Titano version of self test code

gmeader:
- **Titano version of self test code, supports HX8357 320x480 display. Touch screen coordinates corrected so that Sound button-press works**


ladyada:
hi thanks :) it needs to go in its own folder, you cannot have two ino in one spot :)

gmeader:
OK

PR #1260, opened 2020-10-02
first commit

jedgarpark:
first commit of pixel art slideshow code for Matrix Portal, example .bmp files


Pixel_Art_Slideshow/pixel_art_slideshow.py
```diff
@@ -0,0 +1,66 @@
...
+    dwell=IMAGE_DURATION,
+    h_align=HorizontalAlignment.RIGHT,
+)
+last_advance = time.monotonic()
+
+
+def advance():
+    # pylint: disable=global-statement
+    global align_right, last_advance
+    align_right = not align_right
```
makermelissa, line 48: I'm not sure if you want all images to alternate or only 32. If you only want 32 pixel width images to change this, you could load the image first, get the image with slideshow. current_image_name(), load it up, check the width, and then flip align_right only if it's less than 64. Kinda cumbersome, but slideshow doesn't seem to expose much. Or if you're happy with how it works now, then I'm good with it too.

makermelissa: When I wrote this, I didn't realize it was going to be a mix of 32 and 64 pixel wide images.

jedgarpark: happy with how it works now, thanks.

makermelissa (approved):
Looks good to me.

Pixel_Art_Slideshow/pixel_art_slideshow.py
makermelissa: Thanks

PR #1270, opened 2020-10-16
 Added BrainCraft Google Assistant Script

makermelissa:


BrainCraft_Google_Assistant/gv_buttontotalk.py
```diff
@@ -0,0 +1,589 @@
...
+    board.D6,
+    board.D5,
+    3,
+    auto_write=True,
+    brightness=0.2,
+    pixel_order=adafruit_dotstar.BGR,
+)
+dots.fill(0x000000)
+
+# Setup OLED
```
ladyada, line 74: remove?

ladyada (changes requested):

BrainCraft_Google_Assistant/gv_buttontotalk.py
makermelissa: Oh, good catch.

ladyada (approved):

PR #1271, opened 2020-10-16
Code for QT Py Bracelet tutorial

TheKitty:

  the code.

  process.

### DIFF
--- a/QT_Py_Bracelet/code.py
+++ b/QT_Py_Bracelet/code.py
@@ -1,21 +1,15 @@
 # Demo code to generate an alternating color-gradient effect in
-# the QT Py LED cuff bracelet
+# the QT Py LED cuff bracelet LEDs.
 import time
 import board
-import neopixel_spi as neopixel
- 
+import neopixel
+
 # Total number of LEDs on both strips
 NUM_PIXELS = 14
- 
-# Using the neopixel_spi library because the LED strip signal uses the SPI bus. It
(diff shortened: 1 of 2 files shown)


QT_Py_Bracelet/code.py
```diff
@@ -0,0 +1,63 @@
...
+import time
+import board
+import neopixel_spi as neopixel
+ 
+# Total number of LEDs on both strips
+NUM_PIXELS = 14
+ 
+# Using the neopixel_spi library because the LED strip signal uses the SPI bus. It
+#  would work just as well with the standard neopixel library. This particular code
+#  As written, this code doesnt require the faster speed of SPI, but it may prove useful
```
tannewt, line 12: NeoPixel_SPI doesn't provide any extra speed because the transmission still takes the same amount of time as normal NeoPixel.

It is used in cases where we can't reliably bitbang the neopixel protocol, such as Linux. In those cases, we typically *only* have a SPI output that has reliable timing. The Neopixel_SPI library then [converts each neopixel bit to a SPI byte](https://github.com/adafruit/Adafruit_CircuitPython_NeoPixel_SPI/blob/master/neopixel_spi.py#L140) that produces the same output. It's cost is the increased memory per bit.

So, here, just use normal NeoPixel.

tannewt (changes requested):

tannewt (approved):
Looks good! Thank you!

PR #1279, opened 2020-10-20
Adding activity timer and hydration reminder code.

kattni:


QT_Py_Timer/QT_Py_Hydration_Reminder.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+    Convert minutes to seconds for use with time.monotonic().
+
+    :param minutes: The number of minutes. Can be a float, e.g. 0.5 for half a minute.
+    """
+    seconds = minutes * 60
+    return seconds
+
+
+def gradient(color_one, color_two, blend_weight):
```
ladyada, line 54: fancyled has a function like this if you can include that library

ladyada (approved):

QT_Py_Timer/QT_Py_Hydration_Reminder.py
kattni: Good to know. However, the current libs barely fit, another one would not.

PR #1284, opened 2020-10-23
Adding old CPC fruit drum code

caternuson:
This code predates steel, so you can't get it wet...

Doing some guide cleanup. Adding this old CPC code here so can change guide to have embedded links.


ladyada:
hihi needs `.cpx_ada.test.only` files :)

ladyada:
oops - ya need one in each folder

caternuson:
Just for the record - all sketches were re-tested and worked, including the hydrogen MIDI example.

ladyada (approved):

PR #1292, opened 2020-10-30
Pyruler video panic hid fix

FoamyGuy:
I checked all of the projects with `git grep` to find ones using the old HID API, this pyruler video panic button is the last one remaining. 

I believe these changes should fix it on this device, but it is untested. I do not have the pyruler hardware.


ladyada (approved):

ladyada:
good enough for me :)

PR #1295, opened 2020-11-04
fix BOOST_STANDBY pin handling

jepler:
.. because I was testing on a board _without_ `BOOST_STANDBY` I did not initially notice this problem.


FeatherCAN_CircuitPython/listener.py
```diff
@@ -11,7 +11,7 @@
 
 # If the CAN transceiver is powered by a boost converter, turn on its supply
 if hasattr(board, 'BOOST_ENABLE'):
-    standby = digitalio.DigitalInOut(board.CAN_STANDBY)
```
ladyada, line 14: probably best to name variable boost_enable then?

ladyada (changes requested):

ladyada (approved):

PR #1298, opened 2020-11-12
Arduino Esp32-s2 WiFi Examples

brentru:
WiFi test examples from [Adafruit's WiFiNINA library fork](https://github.com/adafruit/WiFiNINA/tree/master/examples) modified for ESP32-S2. Tested with an Adafruit Metro ESP32-S2.


brentru:
Added ESP32-S2 CircuitPython WiFi test example.

ladyada (approved):

PR #1302, opened 2020-11-14
Add quotes example for magtag

ladyada:
two text areas with different fonts with simple text wrapping.


jepler (approved):
thoughts I had while reviewing
 - quote_transform() appears to be unused
 - the resolution of time.monotonic() vs 60 seconds is fine for any plausible scenario

ladyada:
removed the unused transform (was testin stuff)

jepler (approved):

PR #1303, opened 2020-11-14
first commit

jedgarpark:
Magtag slideshow as well as some sample images. The two illustrations are both in the public domain


ladyada (changes requested):
we shoudl probably refactor this example to use the magtag library... do you wanna try to do that?

jedgarpark:
Sure thing, I'll give that a go @ladyada

makermelissa:
It's similar to the matrixportal library

ladyada (approved):

PR #1305, opened 2020-11-15
fix leading zero in minutes for bigclock

ladyada:
no longer 8:3


FoamyGuy (approved):
Tested successfully with `Australia/Eucla` because of course there is a place with a 45 minute offset timezone.

jepler (approved):
lgtm

PR #1306, opened 2020-11-15
 MagTag_CountdownCelebration - fix text position 

jerryneedell:
recent changes to libraries caused some of the text to be shifted off the screen.

changed setting to recenter on the left side.
also ran "black" -- resulted in reformatting several lines


FoamyGuy (approved):
These changes looks good to me. Tested on a device and everything shows nicely on the screen

PR #1307, opened 2020-11-15
Fix date time parsing + Y offset

bergdahl:
Date parsing failed for time zones with positive offsets, like Europe/Stockholm. 
I rewrote the parsing to use substrings instead of split. Also, the text was off screen in the Y direction, so I moved it down to be visible.

### DIFF
--- a/MagTag_CountdownCelebration/code.py
+++ b/MagTag_CountdownCelebration/code.py
@@ -27,7 +27,7 @@ event_time = time.struct_time(
 )  # we dont know day of week/year or DST
 
 # Set up where we'll be fetching data from
-# Chekc http://worldtimeapi.org/timezones for valid values
+# Check http://worldtimeapi.org/timezones for valid values
 # pylint: disable=line-too-long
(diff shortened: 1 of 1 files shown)


ladyada:
hihi can you check, there's a new version that got merged first (with text update) which makes this have conflicts. you may want to 'start fresh'

bergdahl:
I merged the changes, do I need to make a new pull request? And yes, I am new to this. :-)

ladyada:
nope - just need someone to test em! :)

FoamyGuy (approved):
This looks good to me. The old code broke with `Europe/Stockholm` due to the `+` near the end. 
The new code ignores the seconds decimal value and no longer breaks with timezones that are `+` from UTC. It uses index within the string to break it down into components.

Tested successfully on a device, `Europe/Stockholm` now shows correctly instead of raising error.

Thanks for working on this @bergdahl!

ladyada:
thanks for testing + fixing -

PR #1333, opened 2020-12-02
Magtag progress display udpates

FoamyGuy:
The delay times have been increased to one day per fetch. And a note added about using deep sleep once it's available.

The images in the learn guide have been replaced with versions that are cropped further out as well, and the main thumbnail switched.


ladyada (approved):

ladyada:
thanx!

PR #1334, opened 2020-12-02
light cleanup

ladyada:
@FoamyGuy removed some unused variables, etc. please try


FoamyGuy (approved):
I tested both of these successfully. Thank you for these improvements. The weblate one now renders all of text at once instead of a single label at a time which looks much nicer.

I noticed the `time.sleep(4)` was removed from the beginning of the year progress, but is still present in the weblate one. Possibly it should be removed from that one too?

I noticed that having this early in the scripts seemed to alleviate "Display refresh too soon" and occasional random resets while copying files to the device. I know there are many improvements going on in the core for ESP32S2 as well though, so I'm thinking this will eventually get resolved there (if its not already, I'm still on a few days old firmware).

FoamyGuy (approved):
I successfully tested the latest revisions.

Everything is looking good to me. Thank You!

PR #1335, opened 2020-12-02
re-add deleted files

ladyada:


TheKitty (approved):
Reviewed filenames, appears correct

PR #1338, opened 2020-12-02
Adding Cheerlights LED Animation code.

kattni:


jerryneedell:
@kattni I just tried it with a few colors and it worked great! Thanks for this.  So far just got REB MAGENTA and OLD_LACE.
I really like Cheerlights! ooh just went to flashing CYAN.
It is too bad it has to "freeze" while it is updating but I guess that is unavoidable for the animations. I wonder if it would be better to do a fill of the pixels with the last color doing the updates. That way they won't be partially lit from a chase or possible off when flashing. Not a big deal.

jerryneedell (approved):
Tested on MagTag -- looks good to me!

PR #1341, opened 2020-12-03
Add MagTag Tides code

caternuson:
Code and assets.


ladyada (approved):
approved, @makermelissa is gonna update the magtag library to add sleep function so can call that instead of alarm (to abstract away any updates to the api later)

caternuson:
Just in case this matters at some point. Above was run with following firmware build:
```python
Adafruit CircuitPython 6.1.0-beta.1-173-g7ed6e1b5f on 2020-11-28; MagTag with ESP32S2
>>> 
```

PR #1347, opened 2020-12-05
Deepsleep progress displays

FoamyGuy:
tested with:
```
Adafruit CircuitPython 6.1.0-beta.2 on 2020-12-03; MagTag with ESP32S2
```
And the latest MagTag library from release page (should be in the bundle tomorrow I think).

This adds deep sleep to the MagTag progress displays using `exit_and_deep_sleep()`


ladyada (approved):
sounds good!

ladyada:
i updated the guide to have more mirrored pages on setup

FoamyGuy:
Nice look great! I'll add a link or mirrored page to the sleep guide once it's published as well.

PR #1356, opened 2020-12-08
Remove forever loop from MagTag Tides w/ deep sleep

caternuson:
The existing version works fine, but the forever doesn't really ever loop since code just reruns on wake. This PR removes the loop. Tested overnight and works as expected.


ladyada (approved):
thank you!

PR #1357, opened 2020-12-09
Update  MagTag_Christmas_Countdown/code.py to refresh screen after midnight

d-c-d:
sleep till after midnight, so the screen will update about 10 seconds after midnight )

also - catch communication error and restart app if error ( I don't think I'm the only one that sees this )


d-c-d:
edited description for clarity

d-c-d:
Just wanted to follow up - on further testing I am getting an error that I will investigate further and reply here

BlitzCityDIY:
okay! i just loaded it and was going to let it run overnight. my reasoning for 12 hours was that it would update with the proper date almost no matter when someone would start-up the code and looked straight forward if a newer programmer were playing around with it but updating around midnight also makes sense.

d-c-d:
nice to wake up early and see the screen updated for the next day, and it's already sleeping again

BlitzCityDIY (approved):
Ran newest version overnight and runs well updating at midnight, no errors.

d-c-d:
I was eager to check the output this morning too- and it was nice to have the seconds since midnight in the printout- by the time we fetch the clock it reports 30 seconds since midnight ( and we slept till 10 seconds after )

I may look into this further - as others may benefit from knowing accurate times across low power sleeps

PR #1359, opened 2020-12-11
Update covidtracker to pcf font for faster updates

ladyada:
zoom


jepler (approved):
looks good, no testing performed.

PR #1363, opened 2020-12-12
Raspberry pi braincraft radio

jepler:
This adds the script for an upcoming guide.  I accidentally included an unrelated commit which adds comments to an earlier canbus guide for users with esp32s2 microcontroller boards, but this change is good too :)


ladyada (approved):
yep!

PR #1368, opened 2020-12-14
Add MagTag Weather

caternuson:
Code and assets for MagTag weather display.


MagTag_Weather/magtag_weather.py
```diff
@@ -0,0 +1,272 @@
...
+    "October",
+    "November",
+    "December",
+)
+magtag = MagTag()
+
+# ----------------------------
+# Backgrounnd bitmap
+# ----------------------------
+bg_bmp, bg_pal = adafruit_imageload.load(BACKGROUND_BMP)
```
ladyada, line 40: magtag.set_background doesnt work here?

ladyada (approved):

MagTag_Weather/magtag_weather.py
caternuson: Let me look. Mainly took that approach for more z ordering control of background along with other items.

ladyada: no biggie, i think we sorta overrode the magtag graphics system here, which is ok!

caternuson: Ok, that'll work here. Left over thing from tides, where "background" was more of an intermediate layer thing.

PR #1378, opened 2020-12-20
Avoid two refresh in a raw.

dglaude:
The current code does a first refresh when "Year Progress:" is displayed, then after the date is recovered another refresh take place in the final set_text.
By disabling the auto_refresh of the first set_text, it avoid those two refresh... but it also slow down the display of any information at all.

I prefer the effect with this PR, but I can understand the suspense introduced by the original version.


dglaude:
@FoamyGuy please give your opinion on this behaviour.
It took me a bit of time to understand how and when refresh take place.
PS: Feel free to discard.

dherrada (approved):
Looks good. tested

dherrada:
Will merge when passing

PR #1387, opened 2020-12-27
less blasty output

ladyada:
give us a break :)


dherrada (approved):

dherrada:
Will update the guide to reflect the change

PR #1389, opened 2020-12-28
Added links and removed pin list

dherrada:


TheKitty (approved):

dglaude:
Thank you for the credit and long discussion with @dherrada on the how to best do what this code is doing, and more importantly what to do in the other guide to optimise for size (like sharing fonts, and using binary fonts).

dherrada:
@dglaude glad to give credit where credit is due. Yeah, I definitely felt that conversation was really good and hopefully we'll be able to make some stuff more standardized in the future.

PR #1396, opened 2021-01-01
change to new audiocore imports

FoamyGuy:
This updates some imports to use the new ones from `audiocore` instead of `audioio`

Untested, don't have hardware for it.


jerryneedell (approved):
Tested with MAX98357 on a "custom" build of QTPy Haxpress that includes audiobuiso.
Both example work with these changes.

PR #1398, opened 2021-01-04
Fix MagTag Twitter Example

brentru:
Fixes `Http Error: code 401: Authorization Required` caused by multi-line URL split .


ladyada (approved):
sneaky!

PR #1400, opened 2021-01-06
lofi radio: Make control stick follow direction of rotation

jepler:
@djecken Please check out this Pull Request and let me know if it works for you.  It updates the lofi radio scripts so that the control stick's directions follow the screen's rotation automatically.

As always, make sure your Pi is out of Overlay mode when making updates to it, otherwise the changes don't stick.

Grab setup-kiosk.sh from this PR and run it to update.  Then, you can follow the "[kernel module install](https://learn.adafruit.com/adafruit-braincraft-hat-easy-machine-learning-for-raspberry-pi/display-kernel-module-install#installing-the-1-dot-54-kernel-module-3071768-4)" instructions but select any of 0/90/180/270 degrees rotation, remembering that the default was 90:
```
sudo python3 adafruit-pitft.py --display=st7789_240x240 --rotation=90 --install-type=fbcp
```

### DIFF
--- a/Raspberry_Pi_Braincraft_Radio/setup-kiosk.sh
+++ b/Raspberry_Pi_Braincraft_Radio/setup-kiosk.sh
@@ -61,6 +61,15 @@ import os
 import sys
 from datetime import datetime
 
+# If you change the LCD rotation, also change this line:
+ROTATION=90
+
+print("Using rotation", ROTATION)
+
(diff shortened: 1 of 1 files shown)


ladyada:
is this good to merge now?

jepler:
#1402 did this better, so closing this up ...

PR #1402, opened 2021-01-07
Lofi radio: Add automatic rotation fix

makermelissa:
This reads the important values from /boot/config.txt to determine the rotation and then uses a deque to rotate the arrow keys easily based on the detected rotation.


jepler (approved):
Tested at 90 and 270

PR #1422, opened 2021-01-20
remove deprecated button api use

FoamyGuy:
Updates the PyPortal User Interface code to use the new display_button API which gets rid of the warnings printed.


ladyada (approved):
not tested, i trust ya tested it!

FoamyGuy:
Yep, I tested with:
```
Adafruit CircuitPython 6.1.0-rc.1 on 2021-01-15; Adafruit PyPortal with samd51j20
```

PR #1425, opened 2021-01-24
Update code.py

twistedvein:
Fixed the import statements so there's no error.  Changed the pixel number to 44 as called for in the project.


djecken (commented):
Thanks! Tested code on following hardware and version of circuit python:

Adafruit CircuitPython 6.1.0 on 2021-01-21; Adafruit ItsyBitsy nRF52840 Express with nRF52840

ladyada:
@djecken ok plz 'approve' it by clicking on 'files changed' and then approve

djecken (approved):

PR #1432, opened 2021-01-28
CircuitPython_RGBMatrix: Add "tiled flag of wales" example

jepler:
This code is actually by @PaintYourDragon, thanks!


FoamyGuy (approved):
Code and bmp file look good to me. Great to see the tiling made easier! 

Did not test though. I only have enough matricies for a 2x1, and wouldn't be able dig them out and test until later on tonight.

PR #1434, opened 2021-01-29
Added Home Assistant Sensor project code

makermelissa:
Code for my Home Assistant guide

### DIFF
--- a/Home_Assistant_Sensor/code.py
+++ b/Home_Assistant_Sensor/code.py
@@ -1,0 +1,69 @@
+# adafruit_minimqtt usage with native wifi
+import time
+import board
+import busio
+import ssl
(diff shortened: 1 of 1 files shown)


ladyada (approved):

ladyada:
maybe makes sense to go into deep sleep when done? i think people would like that!

makermelissa:
Oh right, I forgot it could do that. I'll see if I can get that working.

makermelissa:
I modified it for deep sleep and am running into this bug: https://github.com/adafruit/circuitpython/issues/4079 so it basically stops working after waking up.

makermelissa:
I decided to give users the option for when it does get fixed.

PR #1449, opened 2021-02-10
Adding and updating data logging examples.

kattni:
Added comment to original data logging example, and added examples for Pico guide.


dhalbert (approved):
Looks good!

PR #1452, opened 2021-02-15
dc motor code

jedgarpark:
Pico motor party example code. DC motor for now, adding stepper, servo, and solenoid next.


ladyada:
plz put in the motor library

jedgarpark:
@ladyada will do -- should I leave this here for the solenoid code? That doesn't use adafruit_motor.

ladyada:
yah you can PR the solenoid code, id close this PR and make a new one

PR #1453, opened 2021-02-15
solenoid demo for Pico

jedgarpark:
Simple solenoid demo using digital pin, MOSFET on RP2040 Pico


ladyada:
@jedgarpark i think the blink() fucntion is a little confusing? if you only blink the led once why not just light up the led while the solenoid is striking?

jedgarpark:
@ladyada I have the LED turn on at start so people know the board is on and running, then I blink when the button is pressed.

ladyada:
oh ok then

ladyada (approved):

PR #1457, opened 2021-02-16
Adding shorter boot.py

kattni:
Added section at the end that covers removing/renaming boot.py from the REPL to allow for data logging without a ground wire.


dhalbert (approved):
Looks good, in the context of the guide.

PR #1463, opened 2021-02-19
Now uses datetime library

jedgarpark:
Uses the datetime library which fixes an issue with timezone conversion getting the date wrong.


jepler (approved):
A nice improvement! (no testing performed)

PR #1474, opened 2021-03-03
workflows: Use 18.04 for arduino-ci

jepler:
The workflow doesn't ask for a specific version of Ubuntu, but when it gets 20.04, ci-arduino causes a failure because it requests installation of a specific version of llvm that is not available in 20.04:
```
E: Version '1:8-3~ubuntu18.04.2' for 'libllvm8' was not found
```

We should decide between this and actually fixing arduino-ci.


dhalbert (approved):
Let's do this to fix the immediate problem, and then research bringing the `clang-format` stuff up to date.

PR #1494, opened 2021-03-27
reduce memory usage in ez make oven

FoamyGuy:
I've made a few changes in the EZ Make Oven code to try to conserve memory in order to hopefully allow it run on a PyPortal Titano as is discussed in this forum post: https://forums.adafruit.com/viewtopic.php?f=60&t=177110&p=863371#p863371

I did successfully replicate the same memory error on the regular PyPortal by hardcoding the Titano's screen size.

The primary things I changed are: 

- change `label` to `bitmap_label`
- remove the calls to `load_glyphs` on all 3 fonts. These are good for speeding up changes to the text that is showing by pre-loading all of the possibly used glyphs but do consume some memory. This app was not using the full set of glyphs so many of the pre-loaded ones were using space with no real benefit. There is ongoing work in the core bitmap.readinto that will get utilized by bitmap_font which will speed this up some. There is also the possibility to convert these font files to pcf which could further speed up the loading time.
- change to using `display_group` instead of `pyportal.splash` I'm not sure if this really made that big of a difference. Originally I was thinking of removing the PyPortal library altogether, and I did experiment some with that but it did not yield much savings seemingly so I put it back.

I also ran black code format which changed several other lines, but not functionally.

I tested this on a regular PyPortal both "as-is" and modified to hard code the Titano's screen size of 480x320. In both cases I am able to run this script successfully after the changes noted above. Whereas without these changes I did get the memory error when I ran with the 480x320 size. I did not have a temperature sensor (or anything else) connected to the pyportal during testing, I'm unsure if that would make a difference.

We did have another community member on discord try running this on a PyPortal Titano and they reported that it was able to successfully run with these changes. They also did not have the sensor or anything else hooked up.

### DIFF
--- a/PyPortal_EZ_Make_Oven/code.py
+++ b/PyPortal_EZ_Make_Oven/code.py
@@ -2,6 +2,7 @@ import time
 import json
 import array
 import math
+import gc
 import board
 import busio
 import audiocore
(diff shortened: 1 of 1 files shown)


ladyada:
oki thanks, i suspected it was something with the labels changing!

ladyada (approved):

PR #1495, opened 2021-03-27
Remove the dependency on PyPortal in the PyPortal_EZ_Make_Oven learn …

lesamouraipourpre:
Inspired by @foamyguy 's Friday stream and the suggestions made in chat, I've removed the dependency on PyPortal as only some audio code was being used. This code has been copied into the Beep class.

Tested on a PyPortal Pynt

Before:

Screen size: 320x240
Mem before imports: 185312
Mem after imports: 61408
WxH: 320x240
Mem after gc before main loop: 18672

Screen size: 400x300
Mem before imports: 185280
Mem after imports: 60448
WxH: 400x300
Mem after gc before main loop: 9872
  File "<stdin>", line 649, in <module>
  File "/lib/adafruit_display_text/__init__.py", line 379, in text
  File "/lib/adafruit_display_text/bitmap_label.py", line 561, in _set_text
  File "/lib/adafruit_display_text/bitmap_label.py", line 237, in _reset_text
MemoryError: memory allocation failed, allocating 616 bytes

After:

Screen size: 320x240
Mem before imports: 184896
Mem after imports: 142528
WxH: 320x240
Mem after gc before main loop: 103808

Screen size: 400x300
Mem before imports: 184896
Mem after imports: 142528
WxH: 400x300
Mem after gc before main loop: 96048

Screen size: 800x600
Mem before imports: 184896
Mem after imports: 142528
WxH: 800x600
Mem after gc before main loop: 21248

### DIFF
--- a/PyPortal_EZ_Make_Oven/code.py
+++ b/PyPortal_EZ_Make_Oven/code.py
@@ -17,7 +17,7 @@ import adafruit_touchscreen
 from adafruit_mcp9600 import MCP9600
 
 TITLE = "EZ Make Oven Controller"
-VERSION = "1.3.0"
+VERSION = "1.3.1"
 
 print(TITLE, "version ", VERSION)
 time.sleep(2)
--- a/PyPortal_EZ_Make_Oven/code.py
+++ b/PyPortal_EZ_Make_Oven/code.py
@@ -5,18 +5,16 @@ import math
 import gc
 import board
 import busio
+import audioio
 import audiocore
 import displayio
 import digitalio
-from adafruit_pyportal import PyPortal
 from adafruit_bitmap_font import bitmap_font
 from adafruit_display_text import bitmap_label as label
 from adafruit_display_shapes.circle import Circle
 from adafruit_button import Button
 import adafruit_touchscreen
 from adafruit_mcp9600 import MCP9600
-
-pyportal = PyPortal()
 
 TITLE = "EZ Make Oven Controller"
 VERSION = "1.3.0"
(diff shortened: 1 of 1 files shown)


PyPortal_EZ_Make_Oven/code.py
```diff
@@ -5,19 +5,17 @@
...
 from adafruit_display_text import bitmap_label as label
 from adafruit_display_shapes.circle import Circle
 from adafruit_button import Button
 import adafruit_touchscreen
 from adafruit_mcp9600 import MCP9600
 
-pyportal = PyPortal()
-
 TITLE = "EZ Make Oven Controller"
 VERSION = "1.3.0"
```
FoamyGuy, line 20: We should update the version here.  _(I should have done this too, and didn't unfortunately)_ 

Since no functionality changed I think we can go with `"1.3.1"`

FoamyGuy (changes requested):
Thanks for working on this @lesamouraipourpre!

I tested this on a regular PyPortal and I do see the same memory savings. I also tested the new Beep object and it does seem to be working as intended.

I have small request. But this is looking good to me otherwise.

PyPortal_EZ_Make_Oven/code.py
lesamouraipourpre: Version bumped

FoamyGuy (approved):
Looks good to me. Thanks again!

ladyada:
thanks!

PR #1497, opened 2021-03-29
Fix raising exception

tannewt:
Reported here: https://forums.adafruit.com/viewtopic.php?f=60&t=177252&p=864129#p863442

I didn't test but should be right.


ladyada (approved):

jedgarpark:
Tested, that works and is a much clearer error, thanks @tannewt

jedgarpark (approved):

PR #1504, opened 2021-04-11
first commit

jedgarpark:
Funhouse door alert code


makermelissa (commented):
Rather than saving door_value in an external variable, you could pass a parameter to send_io_data() with the value.

jedgarpark:
@makermelissa the door_value is True or False and AIO want 0 or 1.

makermelissa:
You're already sending door_value via MQTT. My suggestions was to make the function:
```
def send_io_data(door_value):
    funhouse.peripherals.led = True
    print("Sending data to adafruit IO!")
    funhouse.network.mqtt_publish("door", door_value)
    funhouse.peripherals.led = False
```
and then just call
`send_io_data(1)` or `send_io_data(0)`

However, you can convert True or False to 1 or 0 with int().

jedgarpark:
Ah, I see. I'll make that change, thanks @makermelissa

makermelissa (approved):
Looks good to me

PR #1509, opened 2021-04-16
library import fix, renamed to code.py to appease the Bundle Fly

jedgarpark:
fixed the accelerometer import which changed w 6.0 I think. Also renamed the file to code.py so drag-drop of the bundle doesn't require a rename.


ladyada (approved):

ladyada:
BUNDLING LIKE THERES NO TOMROROOOOOW

PR #1525, opened 2021-04-28
first commit FunHouse LIFX code

jedgarpark:
FunHouse + PIR sensor motion trips color change on LIFX bulb


ladyada (approved):

ladyada:
this is good for now but i think the socketpool/requests stuff can likely be simplified by grabbing the funhouse `network` obj - @makermelissa can you advise?

PR #1549, opened 2021-05-20
Reorganized directory

dherrada:


kattni (commented):
Should `feather_code.py` be `code.py` or is it that way for a reason?

dherrada:
@kattni Yeah, it's that way for a reason. That example uses 2 devices so I thought it'd be less confusing if one was named so it was clear it was for the neopixel controller and not the pyportal remote

ladyada (approved):

ladyada:
@dherrada please rebase to fix up CI :)

dherrada:
Uhh I did not do that right lol. Fixing

PR #1550, opened 2021-05-21
pyportal includes

ladyada:


hathach:
I think we can safely close this PR. Note, for one code fit all scenarios we can actually use 

```C
#ifdef USE_TINYUSB
#include <Adafruit_TinyUSB.h>
#endif
````

This will always work regardless of the stack selected by users.

PR #1551, opened 2021-05-21
Moved to sidereal days

dherrada:
~~Basically the earth actually takes 23hrs 56mins and 4.1 seconds to make a full rotation.~~
So basically the Earth takes 23hrs 56mins and 4.1 seconds to make one rotation in respect to distant astronomical objects. To be honest, I don't really understand why this is. I think it's got something to do with solar days being affected by the movement of the Earth around the sun but sidereal days not being affected since all the other stars are really far away from us, but the Wikipedia article just made me really confused.

Regardless, this change should allow the astrophotography tracker to be more accurate at longer exposures.


hathach:
rebase and push update to get ci build fixed. There is no changes to the pr.

ladyada (approved):

PR #1560, opened 2021-05-28
Updated webhook plant code

dherrada:


brentru (approved):
May want to add comments describing what `LOW` and `MIN` do, otherwise LGTM once passing CI

dherrada:
@brentru I do explain that in the guide, but I guess it really couldn't hurt and would make it so people don't have to open the guide to remember where they are and what they do

PR #1577, opened 2021-06-22
using current PortalBase much of the complexity of logging to AIO goe…

slootsky:
The way that PortalBase is now structured, it greatly simplifies writing an aio based logger.  This code takes advantage of that simplicity.


makermelissa (approved):
Looks good, but I didn't test.

TheKitty (approved):

PR #1586, opened 2021-06-26
Update aws_gfx_helper.py

amswak:
PyPortal display was not updating Fahrenheit temperature due to missing .text


ladyada (approved):

ladyada:
nice thanks!

PR #1597, opened 2021-07-06
Macropad Hotkeys: trap import errors, add comments to config files

PaintYourDragon:


Macropad_Hotkeys/code.py
```diff
@@ -6,7 +6,7 @@
 key sequences.
 """
 
-# pylint: disable=import-error, unused-import, too-few-public-methods
+# pylint: disable=import-error, unused-import, too-few-public-methods, bare-except
```
ladyada, line 9: dont use bare-excepts. use the importerror or whatever it is

ladyada (changes requested):

ladyada (approved):

PR #1604, opened 2021-07-10
Pixel shader fix

FoamyGuy:
These changes update all projects that use OnDiskBitmap to use the new pixel_shader method from https://github.com/adafruit/circuitpython/pull/4823

These changes are necessary in order to run the projects on CircuitPython 7.0.0.beta-3 and newer.

Once support for CircuitPython 6.x is dropped these could be amended further to remove the getattr call and instead just directly pass `bitmap.pixel_shader` In the meantime the changes in this PR allow the projects to work on both older versions and newer versions of CircuitPython


jepler (approved):
I skimmed and didn't see any problems with the changes.  However, there are merge conflicts which will need to be addressed, I think because I also merged some PRs to fix the max group size.  This may indicate needing to work more closely with @lesamouraipourpre when both kinds of updates affect a single file. In any case, thank you!

FoamyGuy:
I will resolve the merge conflicts tonight to get this back to passing.

jepler (approved):
Thank you!

Issue #1603, opened 2021-07-10
Remove max_size and max_glyphs from CircuitPython examples


lesamouraipourpre:
The `max_size` parameter on `displayio.Group` is currently ignored and is (hopefully) going to be removed before 7.0.0 final.
The libraries are having this parameter removed from them - https://github.com/adafruit/circuitpython/issues/4959

Edit: I've update the grep to look for `max_glyphs` related to DIsplay_Text Label as well.

This issue is intended as a tracking issue as the Learn Guide code is worked through
`{ grep -r "Group.*max_size" * && grep -r "max_glyphs" * ; } | sed -e "s/\/.*//" | sort | uniq` to give directories
`{ grep -r "Group.*max_size" * && grep -r "max_glyphs" * ; } | grep -v "_JEplayer_" | sort` gives (updated 2021/08/06 14:20 BST)
```
Matrix_On_Air/matrix_on_air.py:    deco_font, color=color[1], text="AIR", max_glyphs=6
Matrix_On_Air/matrix_on_air.py:    deco_font, color=color[3], text="OFF", max_glyphs=6
Matrix_On_Air/matrix_on_air.py:group = displayio.Group(max_size=22)
PyPortal_Calculator/code.py:calc_display = Label(font, text="0", color=BLACK, max_glyphs=MAX_DIGITS)
PyPortal_Calculator/code.py:calc_group = displayio.Group(max_size=25)
PyPortal_Calculator/titano_code.py:calc_display = Label(font, text="0", color=BLACK, max_glyphs=MAX_DIGITS)
PyPortal_Calculator/titano_code.py:calc_group = displayio.Group(max_size=25)
PyPortal_Titano_Weather_Station/openweather_graphics.py:        self.date_text = Label(self.medium_font, max_glyphs=60)
PyPortal_Titano_Weather_Station/openweather_graphics.py:        self.description_text = Label(self.small_font, max_glyphs=60)
PyPortal_Titano_Weather_Station/openweather_graphics.py:                    self.holiday_text = Label(self.medium_font, max_glyphs=60)
PyPortal_Titano_Weather_Station/openweather_graphics.py:        self._icon_group = displayio.Group(max_size=1)
PyPortal_Titano_Weather_Station/openweather_graphics.py:        self.main_text = Label(self.weather_font, max_glyphs=20)
[...]

PR #1605, opened 2021-07-12
AHT20-OLED: Remove max_size usage from displayio.Group

lesamouraipourpre:
Ref: https://github.com/adafruit/Adafruit_Learning_System_Guides/issues/1603

**Question:** Is this directory used? The obvious Learn Guide search for `AHT20` does not use this. It uses the `simpletest` from the library.
https://learn.adafruit.com/adafruit-aht20/python-circuitpython


jepler (approved):
Thanks.

FoamyGuy (approved):
Changes look good to me. 

I looked around in the Learn Guide system and I could not find any guides that were using this directory. I'm not sure if there is a way to search specifically by the embed URLs.

Still worth getting this updated just in-case it is used somewhere though.

PR #1613, opened 2021-07-14
refactored for Macropad library

jedgarpark:
Updated to use adafruit_macropad library for keys, neopixels, rotary encoder, and push encoder. Left midi as adafruit_midi since it does complex things with midi.receive()


kattni (approved):
Looks good! Shortened it up a bit anyway!

jedgarpark:
thanks -- sorry, I somehow did weird things with the review request when I added Kattni it removed Kattni and re-added Limor!

PR #1615, opened 2021-07-15
first commit

jedgarpark:
This is a demo of basic MIDI functionality on Macropad using `adafruit_macropad` library for keys, LEDs, knob, knob switch, display, and MIDI. Sends noteon/noteoff, program change, cc, pitchbend.

### DIFF
--- a/Macropad_MIDI_Tester/code.py
+++ b/Macropad_MIDI_Tester/code.py
@@ -5,7 +5,7 @@
 # Click encoder to switch modes
 # Turn encoder to adjust CC, ProgramChange, or PitchBend
 from adafruit_macropad import MacroPad
-from rainbow import colorwheel
+from rainbowio import colorwheel
 
 CC_NUM = 74  # select your CC number
 
--- a/Macropad_MIDI_Tester/code.py
+++ b/Macropad_MIDI_Tester/code.py
@@ -1,0 +1,93 @@
+# SPDX-FileCopyrightText: 2021 John Park for Adafruit Industries
+# SPDX-License-Identifier: MIT
+# Macropad MIDI Tester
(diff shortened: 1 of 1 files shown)


kattni (changes requested):
This is lovely! I think the thing that should determine its home, though, is whether or not it's destined for a guide. If so, let's merge it here. If not, I think it is better suited to the MacroPad library examples. Let me know!

jedgarpark:
@kattni cool -- how about we put it in the main Macropad guide then. I can write up a page for it.

kattni (approved):
Ooh, that would be brilliant! Please put the page under "CircuitPython Essentials" titled "MacroPad MIDI". We'll title the examples that use the MacroPad library similarly to identify them.

Macropad_MIDI_Tester/code.py
```diff
@@ -0,0 +1,93 @@
+# SPDX-FileCopyrightText: 2021 John Park for Adafruit Industries
+# SPDX-License-Identifier: MIT
+# Macropad MIDI Tester
+# Play MIDI notes with keys
+# Click encoder to switch modes
+# Turn encoder to adjust CC, ProgramChange, or PitchBend
+from adafruit_macropad import MacroPad
+from rainbow import colorwheel
```
kattni, line 8: This has been updated to `rainbowio`. Please update it here. Thanks!

kattni (changes requested):
Woops! One thing below.

Macropad_MIDI_Tester/code.py
jedgarpark: on it

kattni (approved):
Thanks!

PR #1617, opened 2021-07-17
Improvements to M4_Eyes code

PaintYourDragon:
- Adds “microsaccades” — movement is more natural, less gaze-into-your-soul.
- Initial support for rectangular displays (initially center square only) — code now works on PyBadge or PyGamer with suitably-designed config file and eyelid BMPs.
- Startup calculations now happen during the splash screen idle time if present.
- More yield() calls in more places. Seems the rare “clobber the whole board” bug isn’t internal flash, but filesystem corruption. Will this fix it, or at least make it less likely?


ladyada (approved):

ladyada:
@PaintYourDragon after doing a PR, assign someone to do the review when its ready. i did it this time for ya!

PR #1620, opened 2021-07-20
USB NKRO example for MacroPad

jepler:
This example will be added to the guide "Customizing USB Devices in CircuitPython".

Initially entering as a draft so someone with a macropad (and a non-linux PC:) can test it first.


jepler:
I dug out a Windows laptop and loaded the MacroPad firmware on a Pico to test.

ladyada (approved):

PR #1625, opened 2021-07-22
fix colorwheel location

dborne:
Apparently _pixelbuf no longer has colorwheel in it (https://github.com/adafruit/circuitpython/issues/4992). But it does live in the builtin rainbowio lib,

  the code.
Moving colorwheel import to built in rainbowio lib

Simple import change - just makes the code work now

  process.

Just opened my adabox019 and started playing - and the first demo program was broken. Verified this change works on my shiny new macropad.


kattni (approved):
Thanks for the fix! It's on my list but you beat me to it :)

PR #1637, opened 2021-07-24
CLUE Hand Wash Timer: Update for CP7

lesamouraipourpre:
Remove max_size usage with displayio.Group
Remove max_glyphs usage with Display_Text Label
Update OnDiskBitmap usage with filename string for CP7

Ref: #1603

Learn Guide: https://learn.adafruit.com/no-touch-hand-wash-timer-for-cpx-and-clue
Changes Needed:
Page: https://learn.adafruit.com/no-touch-hand-wash-timer-for-cpx-and-clue/clue-hand-wash-timer
Sections:
* Display and Images: Update the OnDiskBitmap code

### DIFF
--- a/CLUE_Hand_Wash_Timer/clue_hand_wash_timer.py
+++ b/CLUE_Hand_Wash_Timer/clue_hand_wash_timer.py
@@ -13,24 +13,46 @@ import displayio
 import pwmio
 
 clue.display.brightness = 0.8
-clue_display = displayio.Group(max_size=4)
+clue_display = displayio.Group()
 
 # draw the background image
-wash_on_file = open("wash_on.bmp", "rb")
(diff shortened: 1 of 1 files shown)


FoamyGuy (approved):
These changes look good to me. Tested successfully with CLUE CircuitPython 6.3.0 and 7.0.0 alpha3

~The guide page linked uses github code embed so when this gets merged it will get updated automatically.~

FoamyGuy:
I see there is a section that you mentioned on the learn guide that is not a github embed. 

I think we can leave it alone though, it's got the updated code that works for both new and older versions already. The only thing missing is the comment. But I think that is okay in this case since this section is explaining the code at a higher level. I think it's okay to omit the comment to keep those snippets more concise.

TheKitty:
Guide changes made

PR #1638, opened 2021-07-24
CLUE I Ching: Update for CP7

lesamouraipourpre:
Remove max_size usage with displayio.Group

Ref: #1603

Learn Guide: https://learn.adafruit.com/clue-i-ching-caster
*No apparent changes needed to learn guide*


FoamyGuy (changes requested):
This change looks good to me, and the specific max_size change did test successfully on a CLUE device.

If you're up to it can you make one more minor change? It's unrelated to `max_size` but I noticed while testing this that the text in this project is not vertically centered as it appears like it should be based on the images in the guide.

On this line: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/d8c3e592dddd7adcdbffaf07791d8f2104a4dd29/CLUE_I_Ching/clue_iching.py#L74

the second value in the anchor_point should be `0.5` instead of `0.0`

As the code is written now the text is shifted too far down:
image

With `0.5` as the second value in the anchor_point it appears in the center matching the images in the guide:
image

FoamyGuy (approved):
Thank you!

TheKitty:
Verified guide ok

PR #1641, opened 2021-07-26
Add starter examples for Windows, distinguish Mac/Win in title

PaintYourDragon:
Essentially the same three (Adobe Photoshop and Illustrator, and system default browser). Additionally, “Mac” and “Win” have been added to the banner label shown on the display, so users have some extra feedback if they’ve installed the wrong files (e.g. if they just drag-and-drop the lot instead of pick and choose).


ladyada (approved):

ladyada:
good idea!

PR #1642, opened 2021-07-27
Remove gamepad use from Pyloton

dhalbert:
Remove use of `were_pressed()`, and therefore `gamepad`, from Pyloton project. Substitute simple button-push monitoring. Not tested, but looks plausible.

This is to accomodate the removal of `gamepad` in CircuitPython 7.0.0.


kattni (approved):
Thanks for the fix!

dherrada:
Yeah, this seems fine. Thanks for fixing!

PR #1664, opened 2021-07-29
Add Firefox Linux and Numpads to Macropad examples

emceeaich:
  the code.

This changeset adds two new macropads to the Adafruit/Digikey macropad examples. One for Firefox on Linux, and a universal num pad (handy for us 40% keyboard fans)

None.

  process.

These numpads have been tested on a Dell XPS running Linux Minit


lesamouraipourpre (commented):
The number grid as arranged is a Phone pad not a number pad (789 should be in the top row).
The colors in `numpad` look like they were copied from the `linux-firefox` layout and should be changed to a consistent color.

emceeaich:
> The number grid as arranged is a Phone pad not a number pad (789 should be in the top row).

Thanks for catching that. 

> The colors in `numpad` look like they were copied from the `linux-firefox` layout and should be changed to a consistent color.

I looked back at the other macros and see what the developer was doing with setting colors by role/group so I updated the colors to follow the pattern.

I also tweaked the colors on the Linux FF keypad.

Thank you!

PaintYourDragon:
Appreciate the additions. Please update the comments to reflect their actual contents (both say Safari for Mac right now). Thanks.

PaintYourDragon (approved):

PR #1670, opened 2021-07-30
Template cleanup.

kattni:
THIS BREAKS A FEW GUIDES. All of them are on my list to fix. I will need to update one template, and then update four guides to match. Otherwise, the rest of this is only links in template descriptions, and therefore does not break anything.


TheKitty (approved):
Changes look good

kattni:
I will merge this after I eat so I can do the serious fixes immediately.

PR #1677, opened 2021-08-01
update number_crunching for ulab renames and other fixes

FoamyGuy:
This PR updates all of the examples in the fast number crunching project to use the newest version of `ulab`

There are also a few other fixes to some of these examples: 

Change `_pixelbuf` to `rainbowio` for the use of `colorwheel` in the `ledwave.py` example

Split the big list of coefficients into multiple smaller lists in `cluebarometricpressure.py`  to fix "Pystack exhausted" error on CLUE

Set the value returned by `numpy.roll()` into the `data` variable in `cluebarometricpressure.py` and `cluepulse.py` to make the filtered data values come out accurately instead of flat horizontal.

I tested all of these modifications on a CLUE with 7.0.0 alpha5 with the exception of `ledwave.py` which I tested on a Feather M4 with 7.0.0 alpha5

There will need to be updates on these guide pages to reflect the renaming that has occurred in `ulab`. Perhaps some rewording of a few of the ideas presented on the pages will need to occur since they make reference to the previously used sub-modules which are now contained mostly inside of `ulab.numpy` instead of separate sub-modules in some cases.:

- https://learn.adafruit.com/ulab-crunch-numbers-fast-with-circuitpython/overview
- https://learn.adafruit.com/ulab-crunch-numbers-fast-with-circuitpython/ulab-numpy-phrasebook

I have not made these guide updates yet, but am willing to work on them.

I believe the rest of the pages will get handled by the github embeds.


TheKitty:
I'll look to review and merge likely Tuesday after I get the newsletter out. I want to update the guide immediately after the merge to have things all set.

jepler (approved):
thank you!

PR #1678, opened 2021-08-01
weather display matrix cp7 updates

FoamyGuy:
Ref: #1603 

Updates for cp7 in the weather display matrix project.

Tested on Matrix Portal 7.0.0 alpha 5.

Guide: https://learn.adafruit.com/weather-display-matrix/overview 
No apparent changes need.

### DIFF
--- a/Weather_Display_Matrix/openweather_graphics.py
+++ b/Weather_Display_Matrix/openweather_graphics.py
@@ -32,7 +32,7 @@ class OpenWeather_Graphics(displayio.Group):
             am_pm=True,
             units="imperial"
     ):
-        super().__init__(max_size=3)
+        super().__init__()
         self.am_pm = am_pm
         if units == "metric":
             self.celsius = True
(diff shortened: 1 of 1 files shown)


Weather_Display_Matrix/openweather_graphics.py
```diff
@@ -76,6 +82,20 @@ def __init__(
             x=0,
             y=0,
         )
+        # # CircuitPython 7+ compatible
+        # icons = displayio.OnDiskBitmap(icon_spritesheet)
+        # self._icon_sprite = displayio.TileGrid(
+        #             icons,
+        #             pixel_shader=icons.pixel_shader,
+        #             width=1,
```
lesamouraipourpre, line 90: `x`, `y`, `width`, `height` & `default_tile` can all be removed as they are the default values.

lesamouraipourpre (changes requested):
Looks good to me.

The `_icon_sprite` creation can be simplified be removing the parameters that match default values.

Weather_Display_Matrix/openweather_graphics.py
FoamyGuy: Thank you, removed them with the latest commit.

TheKitty (approved):

PR #1682, opened 2021-08-01
snowglobe cp7 updates and background fix

FoamyGuy:
updates for cp7. Ref: #1603 

Also noticed an issue with using a color in the background causing an `AttributeError` I suspect the type of error raised by the core code was changed and this type was not one the ones that were caught so it caused the code to crash. Fixed by adding it to the types that are caught so the code will successfully move on and set the background color.

Tested with CircuitPlayground Bluefruit 7.0.0 alpha5 + TFT Gizmo

Guide: https://learn.adafruit.com/circuit-playground-tft-gizmo-snow-globe/overview-2
No changes needed in the guide


lesamouraipourpre (approved):
Looks good to me. Don't have the hardware to test.

TheKitty (approved):

PR #1720, opened 2021-08-03
Delete boot.py

kattni:


jerryneedell (commented):
Does what it says!

kattni:
The instance of this that is being used needs to be moved. Creating another PR with all the necessary changes.

PR #1744, opened 2021-08-05
lowercase variables in macropad scripts; a few minor pylint fixes

dhalbert:
These scripts were coded with upper-case variables due to pylint being used locally with default settings, and complaining about the names. I lower-cased the non-constants and re-ran pylint in the top-level dir, where it picks up the special `.pylintrc` settings.

I had to change the names of a couple of variables to avoid global/local conflicts. Pylint also picked up one `elif` that could be changed to an `if` (line 191 in dragon_drop).

Re-tested the HotKeys examples, and also the Dragon_Drop game. They work. (I'm not very good at the game.)


TheKitty (approved):

PaintYourDragon:
@dhalbert Thank you for the fixes. For future reference - is there A Particular Canonical Style Guide somewhere you’d recommend following? And if any exceptions to that style, are those codified somewhere? (Since pylint has broad “disable” settings but not “please capitalize constants but not global variables,” things like that which can’t be conveyed in the .pylintrc.) And would you suggest I go through prior projects to bring them up to date with this? Because they’re all The Bad Style.

dhalbert:
@PaintYourDragon pylint tries to implement "PEP8": https://www.python.org/dev/peps/pep-0008/. We have added exceptions to that based on canonical CircuitPython usage, but they aren't written down anywhere that I know of. The main things are:

- It's OK to have a script with most of the code at the top level. However, if the script is getting to be more than a page or two, consider breaking it into separate functions that pass data to each other.
- Use `CAPITALS` for genuine constants, and `lowercase` for variables. The .pylintrc will allow this if it's run when you are in the `Adafruit_Learning_System_Guides` folder.
- We also allow some shorter variable names than pylint would normally allow.

If you have some other projects to fix up, do it at your leisure. Or open an issue here with a list of the ones that could use some style editing, and anyone can then do them and submit PR's.

PaintYourDragon:
Shorter variable names is much appreciated. Whenever pylint complains about that, I can’t help but imagine centuries of mathematicians and physicists spinning in their graves. These things have names for a reason! 😄

Could I suggest maybe we do draft a style guide? I was mortified to learn that years of code I’ve been putting into guides has been substandard and makes us look bad, despite following instructions I was given (from someone else, years ago, don’t worry). Lapse of communication, didn’t even know about the top-level .pylintrc until Thursday.

Issue #1747, opened 2021-08-06
Further updates for CircuitPython 7


lesamouraipourpre:
At the bottom is a list of Learn Guides in need of further updates for CircuitPython 7. The changes needed fall into three categories:
- OnDiskBitmap - Update the usage to take a filename string and handle the file opening internally.
- Sprite - Not a CP7 update but some Learn Guides still use `displayio.Sprite` which doesn't exist anymore.
- TileGrid(position=(0,0)) - Not a CP7 update but some Learn Guides still use `position` with `displayio.TileGrid`. This was replaced with `x` and `y`

Learn Guide code to be updated:
- [x] CircuitPython_displayio - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1749
- [x] CircuitPython_RGBMatrix - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1750
- [x] CircuitPython_sdcardio_sdioio - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1753
- [x] HalloWing_Cat_Toy - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1758
- [x] HalloWing_Jump_Scare_Trap - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1759
- [x] Hallowing_Jump_Sound - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1761
- [x] HalloWing_Tour_Guide - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1762
- [x] Kitty_Paw_Keypad - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1766
- [x] Magic_Nine_Ball - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1765
- [x] Pathfinder - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1754
- [x] PyGamer_Improved_Thermal_Camera - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1764
- [x] PyGamer_Thermal_Camera - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1763
- [x] PyPortal_Alarm_Clock - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1757
- [x] PyPortal_AWS_IOT_Planter - https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/1756
[...]

PR #1755, opened 2021-08-06
RBG Matrix: Update for CP7

lesamouraipourpre:
Update OnDiskBitmap code to take a filename string with CP7.

Ref: https://github.com/adafruit/Adafruit_Learning_System_Guides/issues/1747

Learn Guide: *Unknown*
The only guide that I can find that uses a `rbg.bmp` (Ruth Bader Ginsburg) is https://learn.adafruit.com/image-correction-for-rgb-led-matrices but there is no CircuitPython code in that.

@TheKitty for review

### DIFF
--- a/RBG_Matrix/code.py
+++ b/RBG_Matrix/code.py
@@ -46,11 +46,24 @@ g.append(G)
 display.show(g)
 display.auto_refresh = True
 
+
+# CircuitPython 6 & 7 compatible
 bitmap_file = open("/rbg.bmp", "rb")
 # Setup the file as the bitmap data source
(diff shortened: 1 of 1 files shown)


FoamyGuy:
I did find a blog post that links this code here: https://blog.adafruit.com/2020/09/19/adafruit-matrixportal-makes-it-easy-to-create-a-custom-led-sign-rbg/ 

It doesn't get into the code specifics so no update needed there.

There were a few others that I ran across from the other list that had blog posts but not a full guide.

TheKitty (approved):

PR #1783, opened 2021-08-18
Add support for delay in key sequences

henrygab:
## Scope of change

In the key sequence array, the only data types previously allowed were
of type `int` (for keycodes) and type `str` (for literal text to be typed).

This PR adds support for `float` data type.  This is used to indicate a
time delay in seconds.  It imports module `time`.

Initialization code is indented under `if True:` ... allowing it to be collapsed
in editors such as VSCode.

## Limitations

No known limitations.  Examples require support for `float` in key sequences.

## Examples

Multiple examples are included in this PR, focused on Minecraft (Bedrock Edition).
Examples include both simple delays (e.g., in `minecraft-messages.py`), and complex
command sequences (e.g., in `minecraft-pe-equip.py`).


Macropad_Hotkeys/code.py
```diff
@@ -48,51 +49,50 @@ def switch(self):
...
-files = os.listdir(MACRO_FOLDER)
-files.sort()
-for filename in files:
-    if filename.endswith('.py'):
-        try:
-            module = __import__(MACRO_FOLDER + '/' + filename[:-3])
-            apps.append(App(module.app))
-        except (SyntaxError, ImportError, AttributeError, KeyError, NameError,
-                IndexError, TypeError) as err:
+if True:
```
FoamyGuy, line 52: The final decision to include this or not isn't up to me, but my two cents: I'm not really in favor of adding the `if True:` for indention in these projects. 

The primary audience for these learn guide projects tends to be folks relatively new to programming with little experience. They are encouraged by the guides to use Mu editor which does not have the code block folding feature that comes with some more advanced IDEs like vscode and PyCharm, so it's less likely that they'll get the benefit from this. And it may lead to a bit of confusion if they end up wondering why the if condition is used which may appear useless to them if they don't know about the more advanced editors and their code folding feature.

PaintYourDragon: I’m inclined to agree. This is a good skateboard trick and something I’ll keep in mind for future personal projects…but as it’s a specific editor feature and not a general Python feature, best avoided here.

PaintYourDragon (changes requested):
Excellent feature idea. I’d be happy to incorporate this, but would request some changes:
• Comment at the top of mc-hotbar.py still refers to Universal Numpad, not Minecraft.
• DELAY_AFTER_ESCAPE and DELAY_AFTER_COMMAND are only used in mcpe-equip.py and can be removed from the other macro files.
• Please use 'minecraft' rather than 'mc' in the macro filenames so non-Minecraft people can easily tell what these are for when installing, thanks.

Macropad_Hotkeys/code.py
henrygab: <details><summary>Code folding support is pervasive</summary><P/>

Wikipedia shows that code folding functionality is [quite broadly supported](https://en.wikipedia.org/wiki/Code_folding#Software_with_code_folding_capability) by editors.  In that entire list, only four editors don't support code-folding of Python (either by indentation or by language parsing).

Admittedly, MU does not appear to support code folding.  When Notepad++, VSCode, and both VIM and emacs(!) support a feature, I'd suggest it's an expected capability of code editors.
</details>

But, FoamyGuy has a point.  I've added a comment that should reduce potential for confusion.  PaintYourDragon ... just say the word, and I'll un-indent (but will shed some tears).

PaintYourDragon: Rather de-indent, thanks, sorry.

henrygab:
Should I split the macros into a separate PR?  I haven't been able to test them yet, although I've tested the floating point delay support (against valid inputs) on PC.

henrygab:
OK, this is now tested on Minecraft bedrock edition on PC.

henrygab:
@PaintYourDragon -- All requested changes have been done.  Examples have been tested on Win10 version of Minecraft (aka Minecraft Bedrock edition, aka MinecraftPE).

I'd recommend squash-merging this PR.

henrygab:
Enabled workflows in my branch, fixed the pylint failures.

PaintYourDragon (approved):

PR #1795, opened 2021-08-23
Updated PyLeap Demo Code

TrevKnows:
I've updated the code.py file that is used for the PyLeap demo. This was updated so that the pixelbuf library isn't needed to run this demo. This app will only require the neoPixel.mpy file.

### DIFF
--- a/PyLeap_NeoPixel_demo/code.py
+++ b/PyLeap_NeoPixel_demo/code.py
@@ -3,10 +3,27 @@ import board
 import neopixel
 
 pixels = neopixel.NeoPixel(board.NEOPIXEL, 10, brightness=0.2, auto_write=False)
+rainbow_cycle_demo = 1
+
+def wheel(pos):
+            if pos < 0 or pos > 255:
+                return (0, 0, 0)
+            if pos < 85:
+                return (255 - pos * 3, pos * 3, 0)
+            if pos < 170:
+                pos -= 85
+                return (0, 255 - pos * 3, pos * 3)
+            pos -= 170
+            return (pos * 3, 0, 255 - pos * 3)
+
+def rainbow_cycle(wait):
+            for j in range(255):
+                for i in range(10):
+                    rc_index = (i * 256 // 10) + j * 5
+                    pixels[i] = wheel(rc_index & 255)
+                pixels.show()
+                time.sleep(wait)
 
 while True:
-    for j in range(255):
-        for i in range(len(pixels)):
-            rc_index = (i * 256 // len(pixels)) + j
-            pixels[i] = neopixel._pixelbuf.wheel(rc_index)
-        pixels.show()
+            if rainbow_cycle_demo:
+                rainbow_cycle(0.05)
--- a/PyLeap_NeoPixel_demo/code.py
+++ b/PyLeap_NeoPixel_demo/code.py
@@ -5,7 +5,7 @@ import neopixel
 pixels = neopixel.NeoPixel(board.NEOPIXEL, 10, brightness=0.2, auto_write=False)
 rainbow_cycle_demo = 1
 
-def wheel(pos):
+def colorwheel(pos):
             if pos < 0 or pos > 255:
                 return (0, 0, 0)
             if pos < 85:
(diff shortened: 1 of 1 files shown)


PyLeap_NeoPixel_demo/code.py
```diff
@@ -3,10 +3,27 @@
 import neopixel
 
 pixels = neopixel.NeoPixel(board.NEOPIXEL, 10, brightness=0.2, auto_write=False)
+rainbow_cycle_demo = 1
+
+def wheel(pos):
```
kattni, line 8: We've taken to calling this `colorwheel`, as it is a clearer name. I'd really appreciate it if you could update the code to use `colorwheel` instead of `wheel`.
```suggestion
def colorwheel(pos):
```

PyLeap_NeoPixel_demo/code.py
```diff
@@ -3,10 +3,27 @@
...
+                pos -= 85
+                return (0, 255 - pos * 3, pos * 3)
+            pos -= 170
+            return (pos * 3, 0, 255 - pos * 3)
+
+def rainbow_cycle(wait):
+            for j in range(255):
+                for i in range(10):
+                    rc_index = (i * 256 // 10) + j * 5
+                    pixels[i] = wheel(rc_index & 255)
```
kattni, line 23: You'll need to update this as well.
```suggestion
                    pixels[i] = colorwheel(rc_index & 255)
```

kattni (changes requested):
Hey, Trev! Thanks for the update. You're definitely on the right track. I have a couple of questions. Does PyLeap work with CircuitPython 6.3.0, or does it only work with CircuitPython 7.x? And, is there a specific reason you only want to require the NeoPixel library?

I ask because we added `rainbowio` to CircuitPython 7.0.0 which includes the `colorwheel` code. It is available for every board, so there's no concerns with compatibility, and it simplifies the code significantly. We removed `colorwheel` from `_pixelbuf` in CircuitPython 7, and renamed `_pixelbuf` to `adafruit_pixelbuf`, so you're definitely right to remove that from this example. But if it's possible, `rainbowio` would be helpful here.

However, if PyLeap works with CircuitPython 6, then please wait on using `rainbowio` for now, and see to my suggestions below.

kattni:
@TrevKnows To be clear, `rainbowio` is built into CircuitPython 7+, so you don't have to copy a new library to your board to use it, it's simply available if you're using CP7+.

PyLeap_NeoPixel_demo/code.py
TrevKnows: Good point! I'll update this.

PyLeap_NeoPixel_demo/code.py
TrevKnows: @kattni I wasn't aware of rainbowio. Cool. That would make for less file transferring on my end(which is a great thing). Can you point me in the direction of some rainbowio examples or uses for Circuit Playground Bluefruit?

TrevKnows:
@kattni I wasn't aware of rainbowio. Cool. That would make for less file transferring on my end(which is a great thing). Can you point me in the direction of some rainbowio examples or uses for Circuit Playground Bluefruit?

kattni:
@TrevKnows So, `rainbowio` contains only the `colorwheel()` code. You would still need to use `neopixel.mpy`. As well, before suggesting using `rainbowio`, I need to know what version of CircuitPython PyLeap is using. Please let me know that first. Thanks!

TrevKnows:
@kattni Oh, sorry. I was overly hasty to reply. PyLeap uses both CP6 and CP7+. The goal for this demo is to transfer code.py with rainbow wheels code and neopixel.mpy to support it. It would be easier to just update the code.py to work with both CP6 and CP7+, so I updated the code.py to not use _pixelbuf

kattni:
Ok. Sounds good. For now, please update `wheel` to `colorwheel` in the code, and I'll get it merged.

When CircuitPython 7+ is a stable release, we'll revisit this and update it to use `rainbowio`.

TrevKnows (commented):
code.py updated

kattni (approved):

kattni:
Thanks for the update, @TrevKnows! CI failed on an unrelated thing, so I reran it. Let's see if that works!

TrevKnows:
@kattni Ok, I didn't know if I've done something wrong here. Thank you for helping me.

kattni:
@TrevKnows Great job with this! It's all set now!

PR #1796, opened 2021-08-23
PR 12 - Renamed 59 more files

dherrada:


kattni (changes requested):
The MacroPad Olympic Hotkeys files should not be renamed to code.py as they are shortcuts files. I don't think any of them should have been renamed, even the "home" one. Please verify this by checking the guide.

dherrada:
@kattni Oh yeah I think you are right. I've got to head out for a few to go to an appointment but I'll take care of that when I'm back.

TheKitty:
I believe @kattni is correct. The MacroPad HotKeys guide code.py imports the key files I believe.

PR #1806, opened 2021-08-26
Update CSV column index to account for new total_boosters column

joshkehn:
[Vaccination data](https://raw.githubusercontent.com/owid/covid-19-data/master/public/data/vaccinations/country_data/United%20States.csv) in this guide has a new column `total_boosters` at the end. This means the indexing here is off by one and fails to parse. The failure to parse could be handled as well, but this works and is the smallest fix.

  the code.

  process.

### DIFF
--- a/MagTag_Covid_Vaccination/code.py
+++ b/MagTag_Covid_Vaccination/code.py
@@ -58,8 +58,8 @@ try:
     value = magtag.fetch().split("\n")[-2].split(",")
     print("Response is", value)
 
-    vaccinated = int(value[-2]) / 331984513
-    fully_vaccinated = int(value[-1]) / 331984513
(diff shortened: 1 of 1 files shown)


dherrada:
@joshkehn Thanks for the PR! I'm working on a more robust way to do this that doesn't entirely rely on indexes, since this is like the third time this has happened.

dglaude:
Thank you, I was confused while trying the code from the learn guide, that I had to modify to make it work. I was wondering if it was just me or if something changed. I guessed the format had changed. Reading the first line to identify the field would make things more robust.

dherrada:
@dglaude That's exactly what I did. Fix is in #1819. If you could test it that'd be amazing

PR #1815, opened 2021-08-31
Added UART code

dherrada:


TheKitty (approved):

UART-Between-Boards/code.py
```diff
@@ -0,0 +1,56 @@
...
+barray = bytearray(array)
+print(barray)
+
+uart.write(barray)
+while True:
+    data = uart.read(32)  # read up to 32 bytes
+    if data is not None:
+        msg = []
+        print(data)
+        if data[0] == 85 and data[-1] == 170:
```
tannewt, line 29: I'd factor out 0x55 and 0xAA so that you have constants.

tannewt: This check will breakdown if you receive two packets with the read. Do you want to handle if the read occurs in the middle of a packet?

tannewt (commented):
How reliable are you going for? It might be nice to make functions for send and receive. That would make it easier to use for other data.

dherrada:
@tannewt So my understanding is that this is supposed to be a really simple example

UART-Between-Boards/code.py
dherrada: @tannewt By factoring them out, you mean setting them to their own variables, right?

> This check will breakdown if you receive two packets with the read.

So that was a big issue before I added this block, but after I added that block I stopped having those issues. 

> Do you want to handle if the read occurs in the middle of a packet?

Wouldn't checking the first and last byte solve this issue?

tannewt: > @tannewt By factoring them out, you mean setting them to their own variables, right?

Yup, into a constant so that they are defined in one place.

> > This check will breakdown if you receive two packets with the read.
> 
> So that was a big issue before I added this block, but after I added that block I stopped having those issues.

I think you'd want to check the overall length as well. By just checking start and end, then you'll miss the start and end in the middle of the two packets.

> > Do you want to handle if the read occurs in the middle of a packet?
> 
> Wouldn't checking the first and last byte solve this issue?

It'll drop partial packets so if you miss one, then it might be hard to get back aligned and you'll keep dropping them.

dherrada: > I think you'd want to check the overall length as well. By just checking start and end, then you'll miss the start and end in the middle of the two packets.

Ah ok. That's a good way to do it. 

> It'll drop partial packets so if you miss one, then it might be hard to get back aligned and you'll keep dropping them.

Good point. I think a good solution to this could be storing the last packet sent and resending it if that happens.

tannewt: You probably also want your read to align to your packet size. (So 5 instead of 32.) That way busio.UART can buffer any more incoming packets while you process the most recent one.

When you end up in the middle of the packet you need to read byte-by-byte until you find your start value.

Don't get to the point of resending. Reliable transmission is way harder than just noticing a packet was dropped. (TCP/IP does reliable transmission and it's not simple.)

dhalbert:
I'm working with @dherrada to simplify the protocol.

PR #1816, opened 2021-08-31
Create code.py for new guide

TheKitty:
Code for guide https://learn.adafruit.com/scrambled-number-security-keypad/overview

### DIFF
--- a/MacroPad_Scramble_Lock/code.py
+++ b/MacroPad_Scramble_Lock/code.py
@@ -6,11 +6,11 @@ Scramblepad - a random scramble keypad simulation for Adafruit MACROPAD.
 # SPDX-License-Identifier: MIT
 
 import time
+import random
 import board
 from digitalio import DigitalInOut, Direction
 import displayio
 import terminalio
-import random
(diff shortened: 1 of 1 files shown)


MacroPad_Scramble_Lock/code.py
```diff
@@ -0,0 +1,212 @@
...
+
+while True:
+    if state == STATE_RESET:
+        print("Reset state")
+        macropad.keyboard.release_all()
+        password_guess = ""  # Reset password entry
+        state = STATE_CLEAR  # Reset state
+
+    # Check for key presses/releases
+    event = macropad.keys.events.get()
```
kattni, line 132: For future reference, we prefer to call this variable `key_event` because it's clearer.

kattni (approved):
Did not test. Looks good to me. Non-blocking suggestion below.

MacroPad_Scramble_Lock/code.py
TheKitty: Thanks, copypasta from Drag0n's code, noted

PR #1819, opened 2021-09-02
Came up with better way to get values

dherrada:


TheKitty (approved):

dglaude:
Tested and it works for me.

Unrelated to what this is fixing, I would suggest to work on this:
```
    vaccinated = int(value["people_vaccinated"]) / 331984513
    fully_vaccinated = int(value["people_fully_vaccinated"]) / 331984513
```

For usage in other country, one has to change the value 331984513 twice in the middle of the code, in addition to change the DATA_SOURCE. Putting all the needed change at the beginning of the file would make it easier to localize.

Something like:
```
    POPULATION = 331984513
# ...
    vaccinated = int(value["people_vaccinated"]) / POPULATION
    fully_vaccinated = int(value["people_fully_vaccinated"]) / POPULATION
```

dherrada:
@dglaude Thanks for testing. You're totally right about the population variable.

jepler:
Thanks!  @dglaude feel free to open a PR with the improvement you mentioned.

PR #1821, opened 2021-09-03
demo for p5201

ladyada:


hathach (approved):
I don't have hardware, but it look good on the code side. Approved just in case you want to merge it. Anyone with hardware could confirm the functionality on actual hardware.

PR #1824, opened 2021-09-09
Add I2S pin script, fix cap-touch script

kattni:
My Pylint caught the issue with the cap touch script (since it doesn't ignore the skip file), so I fixed it and removed the skip file.

There are more files incoming, but they need to document the I2S pin script, so I need the URL to be valid, so I'm submitting it separately.


dhalbert (approved):
LGTM!

PR #1828, opened 2021-09-10
Simplify WAV example.

kattni:
Per @dhalbert's request, I have simplified the WAV playback code. We can show the pause/resume function in a different template.


dhalbert (approved):
Thanks - I think this covers the beginner use cases.

PR #1836, opened 2021-09-17
fixing directory structure

jedgarpark:
code got renamed and moved into directories during a cleanup. should work now?


kattni (approved):
This should get the Project Bundle going again. Thanks for the fix.

I hope the .skip file applies to PRs as well. I'm not sure we've done a PR with that file yet. If it fails the linter, we'll have to deal with it.

kattni:
It worked properly. Can be merged once all the Arduino stuff passes.

PR #1837, opened 2021-09-20
update boot.py for 7.0.0

jepler:
This program was originally written for an alpha release of 7.0, and the API changed before the general release. Update the code so it works with 7.0.0.


dhalbert (approved):
Changes look correct for the new API.

PR #1838, opened 2021-09-22
Move to rainbowio

kattni:
Moves all `colorwheel` imports from `_pixelbuf` and `adafruit_pypixelbuf` to `rainbowio`. 

Converts _nearly_ all references to `def wheel()` to use `rainbowio`. There were three `wheel()` functions that are RGBW compatible, and I did not want to remove that functionality from the code as it appears to be referenced in comments. I would like to make `rainbowio` support RGBW, but that is a future thing, so I renamed those `wheel()` functions to `colorwheel()` to make a future update more seamless.

### DIFF
--- a/3D_Printed_LED_Microphone_Flag/code.py
+++ b/3D_Printed_LED_Microphone_Flag/code.py
@@ -31,6 +31,7 @@ import time
 
 import board
 import neopixel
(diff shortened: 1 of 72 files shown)


TheKitty (approved):
Oh my, so many changes! Great work!

PR #1844, opened 2021-09-28
Adding code for the RGB Matrix QT audio visualizer

BlitzCityDIY:
Adding code for the RGB Matrix QT audio visualizer. Uses a Feather Sense to sample audio that is visualized with the RGB LEDs with ulab.


djecken:
Tagging @TheKitty ready for review, thanks!

PR #1855, opened 2021-09-30
Did last 20 file renames

dherrada:
Oops meant to say that I added spdx


ladyada:
@TheKitty you ended up commiting into this PR with arduino code. closing since its probably easier for @dherrada to redo the PR!

dherrada:
@ladyada I think I was able to pull it and fix it, but not sure if I am doing the desired formatting for spdx

PR #1874, opened 2021-10-02
Speed up Kitty_Paw_Keypad

dhalbert:
The KItty_Paw_Keypad project (https://learn.adafruit.com/kitty-toe-bean-paw-keypad-color-tct) was very slow to update. The TileGrid position was being updated every time around the loop, instead of just if it changed.

Tested on a breadboard version of the project.

Thanks to Discord user CptIglo for reporting this.

The guide explanation will need to be slightly revised after this is merged.


FoamyGuy (approved):
The new logic makes sense / looks good to me. Did not test (don't have hardware).

dhalbert:
User reports that it's still missing keypresses, which is true, because it just polls. I am rewriting to use `keypad`.

dhalbert:
A further problem has arisen: https://github.com/adafruit/circuitpython/issues/5423

TheKitty:
@dhalbert - should this be merged?

dhalbert:
> @dhalbert - should this be merged?

There are serious speed problems. Holding in abeyance. I will consult with @BlitzCityDIY.

BlitzCityDIY:
yes i can take a look at this

dhalbert:
@BlitzCityDIY Besides the initial fix in this PR, I tried using monochromatic parrot pictures. That was somewhat faster, but still not great. I think the best solution is not to use `OnDiskBitMap` at all, and instead use in-RAM bitmaps, or some kind of quick animation.

BlitzCityDIY:
@dhalbert solid, yeah will do

Kitty_Paw_Keypad/kitty_pad_imageload.py
```diff
@@ -0,0 +1,139 @@
...
+tft_dc = board.D5
+
+display_bus = displayio.FourWire(
+    spi, command=tft_dc, chip_select=tft_cs, reset=board.D6
+)
+
+#  display setup
+display = ST7789(display_bus, width=240, height=240, rowstart=80)
+
+# CircuitPython 6 & 7 compatible
```
dhalbert, line 60: Doesn't have to be CPy 6 compatible any more. You can just support 7.

dhalbert (commented):
Go ahead and replace code.py with your new code, no need to keep the old code once the guide is updated.

Kitty_Paw_Keypad/code.py
```diff
@@ -52,89 +57,62 @@
...
-                p = 0
-        #  if the button is released...
-        if inputs.value and key_states[i] is True:
-            #  update button state
-            key_states[i] = False
-            #  if a midi keyboard...
+                #  send hid keyboard shortcut
+                keyboard.send(ctrl, shortcuts[event.key_number])
+        #  if a key is released
+        if event.released:
```
dhalbert, line 110: Move increment and picture advance to here.
```suggestion
                #  send hid keyboard shortcut
                keyboard.send(ctrl, shortcuts[event.key_number])
            # advance to next parrot pic; wrap around at end.
            p = (p + 1) % 10
            #  update parrot sprite
            parrot0_grid[0] = p        #  if a key is released
        if event.released:
```

Kitty_Paw_Keypad/code.py
```diff
@@ -52,89 +57,62 @@
...
-        last_p = p
-
-    #  iterate through 4 buttons
-    for i in range(4):
-        inputs = keys[i]
-        #  if button is pressed...
-        if not inputs.value and key_states[i] is False:
-            #  tile grid advances by 1 frame
+    #  update parrot sprite
+    parrot0_grid[0] = p
```
dhalbert, line 92: This is changing the grid every time around the loop. It might be a no-op, but maybe moving it inside the if and adjusting the wrapping increment location as well, as below, would be showing good practice.
```suggestion
```

Kitty_Paw_Keypad/code.py
```diff
@@ -52,89 +57,62 @@
...
+
+    #  get keypad inputs
+    event = keys.events.get()
+    if event:
+        #  if a key is pressed..
+        if event.pressed:
+            #  increase sprite index by 1
             p += 1
```
dhalbert, line 100: Move increment to below.
```suggestion
```

Kitty_Paw_Keypad/code.py
```diff
@@ -52,89 +57,62 @@
...
+            #  if a midi keyboard
             if midi_mode:
-                #  send NoteOff for corresponding MIDI note
-                midi.send(NoteOff(midi_notes[i], 120))
+                #  send note off message
+                midi.send(NoteOff(midi_notes[event.key_number], 120))
+    #  if sprite index is at end of tilegrid
+    if p > 9:
+        #  reset the index to 0
+        p = 0
```
dhalbert, line 118: Functionality moved above.
```suggestion
```

dhalbert (commented):
Thanks for the update! Glad keypad is working for you. I have one restructuring suggestion.

Kitty_Paw_Keypad/code.py
BlitzCityDIY: yesss excellent, will do!

PR #1876, opened 2021-10-04
PR -2 Did 10 more projects

dherrada:


adalogger/adalogger.ino
```diff
@@ -1,3 +1,7 @@
+// SPDX-FileCopyrightText: 2019 Brent Rubell for Adafruit Industries
```
ladyada, line 1: i wrote this code

lightbox/code.py
```diff
@@ -1,3 +1,8 @@
+# SPDX-FileCopyrightText: 2019 Kattni Rembor for Adafruit Industries
```
ladyada, line 1: did kattni work on this code?

ladyada (changes requested):

PR #1882, opened 2021-10-05
Add EyeLights_Bluetooth_Scroller project

PaintYourDragon:


PaintYourDragon:
CI seems to be failing on unrelated code. 🤷‍♂️

TheKitty:
@PaintYourDragon sometimes an Arduino test burps. Best to review the offending error. In this case it appeared to be a timeout, which likely indicates it is not a repeatable error, and will likely pass by rerunning the jobs.

TheKitty:
@ladyada - this one recurringly fails in an Arduino timeout on one specific file, apparently a compiler error and not a code error. We could use someone familiar with the Arduino CI to review.

ladyada (approved):

PR #1886, opened 2021-10-07
try re-enable ci for cpb with Temperature_GIF_Player

hathach:
re-enable ci for Temperature_GIF_Player since it should be fixed by https://github.com/adafruit/ci-arduino/pull/110


hathach:
> @hathach this happens like 5% of runs, do you want to add a 'retry' to the ci script on library installations?

I will make an PR for it

PR #1913, opened 2021-11-06
this change prints entire list of mcu pins along with board mappings

apws:


kattni (changes requested):
Thanks for contributing! I feel like this change has the potential to be confusing. This script is included in the CircuitPython Essentials guide, and the section that refers to it is specific to `board`. There is, however a section immediately below it that discusses the `microcontroller` pins. I would rather have this be a second script that I can then include with the relevant section.

If you're up for it, please create a new directory in the `CircuitPython_Essentials/` directory called `MCU_Pin_Map_Script`, and include this as a `code.py` file in that new directory, with this change. 

It looks like you may have submitted this via the GitHub web interface - I think you should still be able to complete the new request via GitHub, but you may need to submit a new PR, which is fine.

dhalbert:
Closing as stalled. Feel free to reopen.

PR #1917, opened 2021-11-10
Lime update

kattni:
Updated so it'll run on Bluefruit as well - thought it might be good for PyLeap.


TheKitty (approved):

TheKitty:
Approved. This code change breaks all the CPX examples in my book also :(

kattni:
Note - when I say "new" import, this has been in place for a year or two at this point. The existing code worked fine (and will continue to work), but with this change it will now work on CPB as well with no changes.

Guide has been updated to reflect changes.

PR #1925, opened 2021-11-17
Update CP I2C Scan Example

caternuson:
An attempt to update code example to show creating I2C bus manually. QT Py 2040 and Pi Pico are two good examples where existing code does not work and lot's of users end up running into this.

### DIFF
--- a/CircuitPython_Essentials/CircuitPython_I2C_Scan/code.py
+++ b/CircuitPython_Essentials/CircuitPython_I2C_Scan/code.py
@@ -1,4 +1,4 @@
-"""CircuitPython Essentials I2C Scan example"""
+"""CircuitPython I2C Device Address Scan"""
 # If you run this and it seems to hang, try manually unlocking
 # your I2C bus from the REPL with
 #  >>> import board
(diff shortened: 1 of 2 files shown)


ladyada (approved):

CircuitPython_Templates/i2c_scan/code.py
```diff
@@ -1,15 +1,29 @@
 """CircuitPython I2C Device Address Scan"""
+# If you run this and it seems to hang, try manually unlocking
+# your I2C bus from the REPL with
+#  >>> import board
+#  >>> board.I2C().unlock()
+
```
kattni, line 6: Please remove this, as it is explained in the guide page associated with this example.

TheKitty (approved):

kattni (changes requested):
Looks good to me. One thing below.

CircuitPython_Templates/i2c_scan/code.py
caternuson: cool. thanks. i wanted to do that anyway. wasn't sure.

CircuitPython_Essentials/CircuitPython_I2C_Scan/code.py
```diff
@@ -1,9 +1,4 @@
 """CircuitPython I2C Device Address Scan"""
-# If you run this and it seems to hang, try manually unlocking
-# your I2C bus from the REPL with
-#  >>> import board
-#  >>> board.I2C().unlock()
-
```
kattni, line 6: Apologies for not being clear. Please leave it in this copy of the code, it is not explained in the Essentials guide, only in the template page.

kattni (changes requested):
One last thing, sorry!

kattni (approved):

kattni:
@caternuson Thanks for adding this!

PR #1929, opened 2021-11-25
HID Keyboard: handle delayed and terminated connections

DietmarSchwertberger:
The example code is used on many projects, but actually it fails on many occasions, e.g. when shutting down and re-starting a PC.

The modifications handle connection problems. They will try until a successful connection is made and when it is dropped, they will re-try.

Unfortunately I don't know which exception types are thrown exactly. So at the moment I have just added `except:`
I could not test exactly this code, as my hardware is different.

Also, it might be a good idea to add example code for `boot.py`:
```
maintenance_mode = ...  # set when e.g. some key is pressed
usb_hid.enable((usb_hid.Device.KEYBOARD,), boot_device=1 if not maintenance_mode else 0)
```

### DIFF
--- a/CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
+++ b/CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
@@ -39,10 +39,11 @@ while True:
         try:
             # The keyboard object!
             keyboard = Keyboard(usb_hid.devices)
-            keyboard_layout = KeyboardLayoutUS(keyboard)  # We're in the US :)
             break  # successfully connected, leave the connection loop
         except:
             time.sleep(1.0)
+
+    keyboard_layout = KeyboardLayoutUS(keyboard)  # We're in the US :)
 
     try:  # catch exceptions due to terminated connection, e.g. when the PC is shut down
         print("Waiting for key pin...")
--- a/CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
+++ b/CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
@@ -18,11 +18,6 @@ key_pin_array = []
 keys_pressed = [Keycode.A, "Hello World!\n"]
 control_key = Keycode.SHIFT
 
-# The keyboard object!
-time.sleep(1)  # Sleep for a bit to avoid a race condition on some systems
-keyboard = Keyboard(usb_hid.devices)
-keyboard_layout = KeyboardLayoutUS(keyboard)  # We're in the US :)
-
 # Make all pin objects inputs with pullups
 for pin in keypress_pins:
     key_pin = digitalio.DigitalInOut(pin)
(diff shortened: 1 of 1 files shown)


DietmarSchwertberger:
@dhalbert : Would you mind looking at this?
You probably can immediately tell which exceptions are thrown when `Keyboard(usb_hid.devices)` or `press/release/send` fail. I would then update the pull request.
With these exception handlers and your `boot_device` modifications, my new ergonomic keyboard is really working well!

dhalbert:
Thanks for the PR. It is a long holiday weekend in the US; I will take a look at this next week, I hope.

CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
```diff
@@ -36,29 +31,46 @@
...
+    # main loops for connection and key presses
+
+    while True:
+        # stay in this loop until a connection is available
+        try:
+            # The keyboard object!
+            keyboard = Keyboard(usb_hid.devices)
+            keyboard_layout = KeyboardLayoutUS(keyboard)  # We're in the US :)
+            break  # successfully connected, leave the connection loop
+        except:
```
deshipu, line 44: this will also cach ``KeyboardInterrupt`` and many other errors you don't want to catch and hide from the user

CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
```diff
@@ -36,29 +31,46 @@
...
+                        keyboard.release_all()  # ..."Release"!
 
-            # Turn off the red LED
-            led.value = False
+                    # Turn off the red LED
+                    led.value = False
 
-    time.sleep(0.01)
+            time.sleep(0.01)
+    except:
```
deshipu, line 74: ditto

CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
```diff
@@ -36,29 +31,46 @@
...
-            print("Pin #%d is grounded." % i)
+    # main loops for connection and key presses
+
+    while True:
+        # stay in this loop until a connection is available
+        try:
+            # The keyboard object!
+            keyboard = Keyboard(usb_hid.devices)
+            keyboard_layout = KeyboardLayoutUS(keyboard)  # We're in the US :)
+            break  # successfully connected, leave the connection loop
```
deshipu, line 43: you probably want this break in an ``else`` clause of the ``try``

CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
DietmarSchwertberger: That's why I asked Dan to review. He probably knows the execption type(s) to catch.

CircuitPython_Essentials/CircuitPython_HID_Keyboard/code.py
DietmarSchwertberger: Personally I prefer the break at this position as it's clearer. I have moved the KeyboardLaoyutUS code out of try/except, though. I think it can't fail.

dhalbert:
This very old PR has lain fallow, sorry. I recently revisited the same issue. Nowadays another, probably simpler, way to check would be to use `supervisor.usb_connected`. We could revise the code here, or close the PR and update sample code in several places, in conjunction with updating Learn Guide text in various places.

DietmarSchwertberger:
Sure, feel free to close this one, probably after posting a link to the new method.

Would be great to have also a method to fix https://github.com/adafruit/circuitpython/issues/5380 .
This is the remaining issue for building reliable keyboards.

dhalbert:
(Finally) the adafruit_hid library now checks for `.usb_connected` and will wait indefinitely for a host to come up. See https://github.com/adafruit/Adafruit_CircuitPython_HID/pull/118, which is released as https://github.com/adafruit/Adafruit_CircuitPython_HID/releases/tag/6.0.0. So I think this PR is now moot.

DietmarSchwertberger:
Well, with the other, merged PR the initial connection will not timeout. So, a delayed connection does not need to be handled any more.

But what happens when the PC shuts down with the USB ports remaining powered or when it goes to standby. Probably the connection will then fail at some point. So the try/except around the main loop is still required, right?

See also https://github.com/adafruit/circuitpython/issues/5380 . This probably needs some controlled disconnection as well. A solution for this would probably remove the need for try/except for terminated connections.

dglaude:
> (Finally) the adafruit_hid library now checks for `.usb_connected` and will wait indefinitely for a host to come up. See [adafruit/Adafruit_CircuitPython_HID#118](https://github.com/adafruit/Adafruit_CircuitPython_HID/pull/118), which is released as https://github.com/adafruit/Adafruit_CircuitPython_HID/releases/tag/6.0.0. So I think this PR is now moot.

I am worried about that change that will likely break my ultimate mouse jiggle.
https://github.com/dglaude/Mouse_Jiggler_Trinket_M0/blob/main/code.py

In my code I actively wait for the USB to be there while blinking an LED to warn about a problem. That is the behavior if you plug your mouse jiggler in a USB charger... it is also what is happening when your computer boot.

To answer @DietmarSchwertberger I am not sure what is happening at the end, when you turn off your operating system, but the USB port is still powered because it is a charging port. In my code I force a reload on error, so when that occurs, I go back to the initial loop of blink warning.

I'll stop complaining here and try to complain on the usb_hid library about that breaking change. :-)

PR #1930, opened 2021-11-25
Add blank screen usable when idle

ViToni:
The MacroPad stays active including screen and lights even when the computer it is attached to, goes to sleep. 
This change provides a blank screen, sets all RGB lights to black and disables the bright background of the title bar when the title is empty (as on the blank screen).


PaintYourDragon (approved):

ViToni:
@PaintYourDragon Could you also merge this PR? Or does someone else need to do the merge?

PR #1950, opened 2021-12-14
print out traceback for errors in Macropad Hotkeys macro files

dhalbert:
The Macropad Hotkeys script silently ignores errors in `macros/*.py`. This makes it more robust, but also makes debugging user-provided macro definitions more difficult.

@Neradoc suggested adding a traceback print (thanks!). The new behavior is that the errors will be printed to serial output, but the script will still run.


TheKitty (approved):

PaintYourDragon:
Nice. Thank you!

PR #1951, opened 2021-12-20
code for qt py acrylic lamp

djecken:
code for qt py acrylic lamp


djecken:
ready for review @TheKitty thank you!

TheKitty (approved):

PR #1959, opened 2022-01-03
Macropad_2FA_TOTP typo

dglaude:
I guess it is a typo.


TheKitty (approved):
That's leet speak but yes we can change.

PR #1967, opened 2022-01-11
Update Trinket_Ultrasonic_Rangefinder.ino

TheKitty:
Error in and operation


TheKitty:
In reviewing the errors thrown from another project, I noted a logic error flagged by the compiler, standard & vs &&. This corrects that.

ladyada (approved):

PR #1969, opened 2022-01-11
Use default I2C address for NeoSlider CircuitPython example

dhalbert:
The default out-of-the-bag I2C address for a NeoSlider is `0x30`, but the CircuitPython example is using `0x31.` Changed to `0x30`.

(Noticed this while testing NeoSlider for a customer problem.)


kattni (approved):
Oof, good catch. Thanks!

PR #1977, opened 2022-01-15
Remove gampadshift from PyBadge_Conference_Badge example

tekktrik:
Removes `gamepadshift` from PyBadge Conference Badge example and replaces it with `keypad` library so it will be viable after `gamepadshift` is removed in CircuitPython 8.0.0

No changes to functionality, intended to function as it does prior to this change

Ran successfully on PyBadge with CircuitPython 7.1.1, previous behavior retained.  Learn Guide will need to be updated accordingly, as it explains the code.

### DIFF
--- a/PyBadge_Conference_Badge/code.py
+++ b/PyBadge_Conference_Badge/code.py
@@ -8,7 +8,6 @@ from math import sqrt, cos, sin, radians
 import board
 from micropython import const
 import displayio
-import digitalio
 import neopixel
 from keypad import ShiftRegisterKeys, Event
 from adafruit_display_shapes.rect import Rect
@@ -46,13 +45,14 @@ neopixels = neopixel.NeoPixel(board.NEOPIXEL, NEOPIXEL_COUNT, brightness=brightn
                               auto_write=False, pixel_order=neopixel.GRB)
 
 latest_event = Event()
-last_press = None
+last_event = Event()
 
 pad = ShiftRegisterKeys(clock=board.BUTTON_CLOCK,
                    data=board.BUTTON_OUT,
                    latch=board.BUTTON_LATCH,
                    key_count=8,
                    value_when_pressed=True,
+                   interval=0.1,
                    max_events=1)
 
 # Make the Display Background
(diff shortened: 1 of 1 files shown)


tekktrik:
If this is worth merging, I'm happy to change the other usages of `gamepadshift` in other examples!

FoamyGuy:
I accidentally left this comment on the wrong PR initially. This is where it belongs though:

It looks like the code from this PR is used for this guide: https://learn.adafruit.com/pybadge-conference-badge-multi-language-unicode-fonts

In the long term it may be worthwhile to extend the [PyBadger Library](https://github.com/adafruit/Adafruit_CircuitPython_PyBadger) to be able to use custom fonts for the conference name badge. I think this would allow the library to used with different languages like is shown in the guide. 

The PyBadger (and all other) libraries were previously updated to remove usages of gamepad.

tekktrik:
Hey @FoamyGuy, it looks like the PyBadger library can create the conference badge with custom fonts:  https://github.com/adafruit/Adafruit_CircuitPython_PyBadger/blob/3492ff778df773162fabe70caea6f26101932b37/adafruit_pybadger/pybadger_base.py#L546-L591.  I was going to create an issue in the PyBadger library for that, but wanted to make sure I understood before doing so, so let me know if I am.

FoamyGuy:
@tekktrik ah nice catch. I think you are right it does seem to already support custom fonts. So theoretically it should be possible to switch this guide over to using the PyBadger library instead of the more manual approach in it now. 

That would also inherently remove the gamepad usage since that change was already made to PyBadger library, and would simplify / reduce code for the project a little bit. 

It may also be worth just merging in this fix for now though to get it back to working and worry about converting the guide project to use the library further down the road some.

makermelissa:
I'm ok with switching over to the PyBadger library eventually, which was based on the code in question, if the following things can be met:
 * I would like the badge to appear the same (using the fonts) so that we can use the same photos.
 * Some of the fonts were quite large, so it may be a really tight squeeze since the PyBadger library has extra things
 * I don't think the PyBadger does the same Neopixel stuff

For now though, I would rather stick with the changes in this PR.

makermelissa (approved):
Since this is my code, I'm good with these changes. I'll update the guide after this is merged.

PR #1979, opened 2022-01-17
adding circuitpython neko cat

FoamyGuy:
Adding the code for a CircuitPython port of the Neko cat. 

Here's a brief video showing some of the animations that it does:

https://user-images.githubusercontent.com/2406189/149822887-bdc71203-b62a-4e38-8d6f-57856c126698.mp4


TheKitty (changes requested):
PyLint Error 
image

FoamyGuy:
This is a curious error. I'm not seeing it when I run pylint locally. I believe I have the same version of pylint as the actions container (2.7.1). So I'm not certain why I don't see this error from pylint when I run it locally. Maybe due to differing python versions. I have 3.8.10 locally and CI is using 3.10.1.

The error itself doesn't make a lot of sense to me either. It seems to think setting a new value to the `x` and `y` properties is actually defining them rather than just updating them. Those weren't defined in my `__init__()` function because they came from the parent class TileGrid, my class is only updating them, not defining them.

I've added a commit that sets `x` and `y` properties to `0` inside of `__init__()` to see if that resolves it.

TheKitty (approved):

PR #1981, opened 2022-01-18
fix esp32 wifi demos

ladyada:
we have to explicitly say we don't want to load a root certificate nowadays!


TheKitty (changes requested):
image

TheKitty (approved):

PR #1986, opened 2022-01-22
Reformatted to reduce each line length.

Pikime:
Max supported line length in Mu (the recommended editor) is 88 charecters. Several lines in the comments section were longer than this. Changed so it may be copy-pasted as directed in tutorial.


kattni (approved):
Thank you for the fix! I have resolved the Pylint issue - it failed on trailing whitespace. For your own information, Pylint will fail if there are spaces at the end of lines, even in docstrings and comments. They can be sneaky! 

I'm approving this, we'll merge it as soon as it's passing.

PR #2002, opened 2022-02-08
Add/update DotStar single LED template code

kattni:
Tested on ItsyBitsy M4.

### DIFF
--- a/CircuitPython_Templates/status_led_one_dotstar_rainbow/code.py
+++ b/CircuitPython_Templates/status_led_one_dotstar_rainbow/code.py
@@ -1,0 +1,25 @@
+# SPDX-FileCopyrightText: 2021 Kattni Rembor for Adafruit Industries
+# SPDX-License-Identifier: Unlicense
+"""
+CircuitPython DotStar red, green, blue, brightness control example.
+"""
+import time
+import board
+from rainbowio import colorwheel
+import adafruit_dotstar
+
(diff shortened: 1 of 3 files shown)


PaintYourDragon:
With a single LED and with auto_write being the default case, would it be OK to replace:
```
        for led in range(1):
             dot_index = (led * 256 // 1) + color_value
             dot[led] = colorwheel(dot_index & 255)
         dot.show()
```
with:
```
        dot[0] = colorwheel(color_value)
```
?

Sometimes there’s Learning Value in showing things the verbose way, so no worries if that was the intent.

kattni:
@PaintYourDragon Ooh, this is good. I don't need it to be verbose. I will need to update the copy to match it, but it means less copy. In this case, it's simply meant to give rainbows, not necessarily explain how rainbows work deep down. To be clear, you're saying the function should look like this? Or am I missing part of your suggestion?
```
def rainbow(delay):
    for color_value in range(255):
        dot[0] = colorwheel(color_value)
        dot.show()
        time.sleep(delay)
```

PaintYourDragon:
Yep. All the math just distills down to nothing in that case and color_value can be passed straight to colorwheel. And the show() call might be redundant.

kattni:
Wait - I just figured out what I was missing. I never set `auto_write=False`. Removing the `show()`!

TheKitty (approved):

PR #2004, opened 2022-02-08
Change IS31FL3741 DisplayIO switch check

gamblor21:
Small change to make the code more consistent with other examples.


kattni (approved):
Thanks for the fix!

Issue #2011, opened 2022-02-16
IS31FL3741_DisplayIO missing font file.


caternuson:
These guide pages:
https://learn.adafruit.com/adafruit-eyelights-led-glasses-and-driver/displayio-message-scroller
https://learn.adafruit.com/adafruit-eyelights-led-glasses-and-driver/displayio-sports-scroller

Both code examples use a `scrolly.bdf` font file that is missing and does not get included when downloading project bundle.

caternuson:
Fixed with #2012. Thanks @gamblor21

PR #2012, opened 2022-02-16
Include font file with examples

gamblor21:
The font file was not included so when "Download project bundle" was selected it was not included.


caternuson:
Just adding for tracking - this is for #2011

TheKitty (approved):

PR #2027, opened 2022-02-19
Fix the button code in the NeoPixie_Dust_Bag

Neradoc:
Fixed the button detection to only act on rising edges.
Removed the debounce code (the loop already has a time.sleep in the animation anyway).

The guide uses a Standalone Momentary Capacitive Touch Sensor Breakout ([product 1374](https://www.adafruit.com/product/1374)). Which is active high. So the input pin is pulled down. And not using `touchio` (not without changes to the guide).

### DIFF
--- a/NeoPixie_Dust_Bag/code.py
+++ b/NeoPixie_Dust_Bag/code.py
@@ -22,13 +22,13 @@ delay_mult = 8  # Randomization multiplier, delay speed of the effect
 pixels = neopixel.NeoPixel(
     neo_pin, pixel_count, brightness=.4, auto_write=False)
 
-oldstate = True  # counting touch sensor button pushes
+oldstate = False  # counting touch sensor button pushes
 showcolor = 0  # color mode for cycling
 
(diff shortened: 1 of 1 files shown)


NeoPixie_Dust_Bag/code.py
```diff
@@ -22,13 +22,13 @@
...
-oldstate = True  # counting touch sensor button pushes
+oldstate = False  # counting touch sensor button pushes
 showcolor = 0  # color mode for cycling
 
-# initialize capacitive touch input
+# initialize external capacitive touch pad (active high)
 button = digitalio.DigitalInOut(touch_pin)
 button.direction = digitalio.Direction.INPUT
-button.pull = digitalio.Pull.UP
+button.pull = digitalio.Pull.DOWN
```
ladyada, line 31: not every chip has a pull down, id rather stick to a pullup so it doesnt float

ladyada (changes requested):

ladyada (approved):

PR #2051, opened 2022-02-23
Add starter code for Arduino TFT template.

kattni:
This code does not compile as-is; it is meant to be a starter where the guide template page author updates the indicated sections to match the board for the guide into which the template page is going. Therefore, there is a `.none.test.only` because CI would fail otherwise. This code will never be user-facing, it will always be updated and resubmitted to Learn separately, and CI would be run on it at that point, before embedding it into a board guide.

### DIFF
--- a/Arduino_Templates/TFT_Graphics_Test_Template/TFT_Graphics_Test_Template.ino
+++ b/Arduino_Templates/TFT_Graphics_Test_Template/TFT_Graphics_Test_Template.ino
@@ -1,0 +1,314 @@
+/**************************************************************************
+  This is an example of using the built in TFT.
+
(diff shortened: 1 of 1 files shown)


ladyada:
do we need this? we have starter graphicstest code for each board already by @makermelissa already

kattni:
@ladyada I don't see CLUE or FunHouse in the lib I grabbed this from. Would they be elsewhere?

ladyada:
check and work with @makermelissa on it please

kattni:
@ladyada Will do.

kattni:
This is being superseded by the example code available in the ST7735/ST7798 library. The template this was destined for will use the board-specific example available in the library instead of adapting this example repeatedly.

PR #2057, opened 2022-02-23
Added SPDX to 30 more files - spdx-35

evaherrada:


TheKitty (changes requested):
The weather files are by Dan Cogliano https://learn.adafruit.com/epaper-weather-station/arduino-setup

TheKitty (approved):

PR #2060, opened 2022-02-24
Added SPDX to 30 more files - spdx-38

evaherrada:


TheKitty (changes requested):
Please change the attribution of the IR code files to Limor

TheKitty (approved):

PR #2071, opened 2022-02-26
add midinote and midicc to hotkeys

albinger:
Added a 'midinote' and 'midicc' option for the macros
Created a midi.py example showing how to use.


PaintYourDragon (commented):
Please don’t remove the SPDX license header from code.py (that might’ve not been intentional, it can happen if the code changed after you checked out).

Rather than a fixed-duration 0.15 second note, it might be more flexible to have NoteOn on keypress and NoteOff on release, BUT also do something like the key press/release sequences where a negative number allows releasing key(s) in the middle of a sequence (with optional floating-point delays in there). This is a bit more complex but would cover both the DAW and playable instrument cases.

PR #2075, opened 2022-02-27
reduced polyphony lag

jedgarpark:
Code was lagging when four notes were played in rapid succession. Added 'while macropad.keys.events:' loop and turned off display auto refresh, instead refreshing once per loop. Note, 5-6 note polyphony still can lag, due probably to hitting the macropad max key event queue. This can probably be solved by using keypad instead of macropad, but that is beyond the scope of the MacroPad main guide examples.


TheKitty (approved):
LGTM

PR #2077, opened 2022-02-28
Webb telescope update

FoamyGuy:
**NOTE:** This new version requires: [#6097](https://github.com/adafruit/circuitpython/pull/6097) from the core circuitpython. This PR here should not be merged until after #6097 in the core has been. The change in root certificates from #6097 is required in order to get successful request from the newer API source.

This change contains a new version of the James Webb Telescope MagTag project. 

The original version sourced is data from a 3rd party API that went offline when the telescope reached it's intended orbit destination.

The new version gets data directly from a nasa.gov served API so there is no longer a 3rd party server in the mix. 

The data available is different than previously, now we have 9 different temperatures and none of the distance or percentage complete type information so I've updated the display to show all of the temperature values and remove everything that is no longer present in the data.

Here is a picture of the new version:
image

### DIFF
--- a/MagTag_Webb_Telescope_Status/code.py
+++ b/MagTag_Webb_Telescope_Status/code.py
@@ -31,7 +31,9 @@ SLEEP_TIME = 60 * 60  # seconds
 # SLEEP_TIME = 60 * 60 * 24 # seconds
 
 # URL to fetch the data from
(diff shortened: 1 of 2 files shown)


TheKitty (approved):
LGTM

PR #2114, opened 2022-03-20
factory demo for kb2040

ladyada:
also add a wifi demo test in a standalone folder


kattni (approved):
Looks good to me! Did not test.

PR #2117, opened 2022-03-21
Update PyPortal User Interface code to support Titano

aioue:
  the code.

Fixes blank screen/barely legible text on Titano. Without these changes the code *appears* to not function on a Titano which is confusing for newbies.

Does not address PyPortal Pynt.

  process.

Tested on Titano.

### DIFF
--- a/PyPortal_User_Interface/code.py
+++ b/PyPortal_User_Interface/code.py
@@ -64,12 +64,20 @@ def set_backlight(val):
     board.DISPLAY.brightness = val
 
 # Set the Backlight
-set_backlight(0.3)
+if board.board_id == "pyportal_titano":
+    # 0.3 brightness does not cause the display to be visible on the Titano
+    set_backlight(1)
+else:
+    set_backlight(0.3)
 
 # Touchscreen setup
 # ------Rotate 270:
-screen_width = 240
-screen_height = 320
+if board.board_id == "pyportal_titano":
+    screen_width = 320
+    screen_height = 480
+else:
+    screen_width = 240
+    screen_height = 320
 ts = adafruit_touchscreen.Touchscreen(board.TOUCH_YD, board.TOUCH_YU,
                                       board.TOUCH_XR, board.TOUCH_XL,
                                       calibration=((5200, 59000),


ladyada:
@RichardA1 take a look!

FoamyGuy (approved):
Looks good to me. Thanks for adding support for Titano @aioue 

Tested successfully on PyPortal Titano

CedarGroveStudios:
@FoamyGuy @aioue Discovered the source of the Titano brightness problem. The LED current controller on the board is being driven by a PWM signal with a frequency that is outside of the linear response range of the controller. Here's the issue:

https://github.com/CedarGroveStudios/Cat/issues/1

Not sure how to post an issue in `displayio.display` in the core for the fix.

dhalbert:
> Not sure how to post an issue in `displayio.display` in the core for the fix

Thanks for figuring it out! Submit an issue to https://github.com/adafruit/circuitpython/issues, or submit a PR to circuitpython.

FoamyGuy:
@CedarGroveStudios interesting. Thanks for hunting that down! An issue for the core can be created here: https://github.com/adafruit/circuitpython/issues

I wonder if there is any practical downside to just lowering the frequency inside the core for all displays.

dhalbert:
> I wonder if there is any practical downside to just lowering the frequency inside the core for all displays.

500 Hz should not have any noticeable flicker. It should be tested on at least some boards. I wonder if a lower frequency can get into audio channels, though, and cause a background tone. Perhaps that was the reason for the high frequency choice.

CedarGroveStudios:
@FoamyGuy Don't think there'll be an issue with other displays since, in most cases, it's just controlling a small array of backlight LEDs. 

I tried it on the TFT FeatherWings with no issues. No flicker, no annoying buzz. I'll give it a shot on some other backlit displays that are hanging around the workbench. The `pwmio` default frequency of 500Hz seems to be the sweet spot.

Fixing it for built-in displays in `displayio.display` is a one-line change, I think. I'll look at breakout board drivers in the library to see if they will need changes.

CedarGroveStudios:
> > I wonder if there is any practical downside to just lowering the frequency inside the core for all displays.
> 
> 500 Hz should not have any noticeable flicker. It should be tested on at least some boards. I wonder if a lower frequency can get into audio channels, though, and cause a background tone. Perhaps that was the reason for the high frequency choice.

Good point. I don't have a way to test the potential interaction on the PyPortal without modifying the core, something that's outside of my wheelhouse.

dhalbert:
If you open an issue with the fix, we can easily make a PR and make a build. You are the right expert to do the audio testing.

CedarGroveStudios:
> If you open an issue with the fix, we can easily make a PR and make a build. You are the right expert to do the audio testing.

I'm not certain that I picked the right parameter in `displayio.display` #119, but will create an issue after today's meeting.

CedarGroveStudios:
submitted as https://github.com/adafruit/circuitpython/issues/6236

PR #2130, opened 2022-03-31
Simplify blink block.

kattni:
I tried this before submitting the code the first time, and it didn't work, but that turned out to be because I was doing it wrong. Returning the `blink` code back to being more understandable.


dhalbert (approved):
looks good!

PR #2135, opened 2022-04-05
Code tweak, GIF upload.

kattni:
Decided using `animation_controls` across all functions made it unclear. Also updates a `buttons` plural to `button` since there is only one in this example.

Tested the ability to host GIFs on GitHub. The asyncio template supports two hardware setups, one using a built-in button, and one using an external button. I wanted the GIF included in the template to at least match the hardware button set up, since we opted to only do one GIF and not redo it for every board guide in which this template is included. The Learn system allows you to use the Media>Remote URL option with a GitHub link (also tested this). So my plan is to include a template area directing the author to choose from one of the included GIF files, depending on how the button is setup for a given board. That way it at least looks right for the button setup, even if the board isn't exactly the same.

### DIFF
--- a/CircuitPython_Templates/asyncio/code.py
+++ b/CircuitPython_Templates/asyncio/code.py
@@ -15,7 +15,7 @@ On the Feather M4 Express, you would wire the external button to A0, and update 
 
 Two: THIS IS ONLY NECESSARY IF THE BUILT-IN BUTTON IS ACTIVE HIGH. For example the built-in
 buttons on the Circuit Playground Bluefruit and the MagTag are active high.
-If your button is ACTIVE HIGH, under "async def monitor_button(button, animation_controls)",
+If your button is ACTIVE HIGH, under "async def monitor_button(button, controls)",
 update the following line of code:
     with keypad.Keys((button,), value_when_pressed=False, pull=True) as key:
 To the following:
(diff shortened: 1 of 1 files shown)


TheKitty (approved):

TheKitty:
@dhalbert - awaiting your review

CircuitPython_Templates/asyncio/code.py
```diff
@@ -76,22 +76,21 @@ async def monitor_button(button, animation_controls):
...
 async def main():
     animation_controls = AnimationControls()
-    buttons_task = asyncio.create_task(monitor_button(button_pin, animation_controls))
+    button_task = asyncio.create_task(monitor_button(button_pin, animation_controls))
     animation_task = asyncio.create_task(rainbow_cycle(animation_controls))
     blink_task = asyncio.create_task(blink(animation_controls))
 
     # This will run forever, because no tasks ever finish.
-    await asyncio.gather(buttons_task, animation_task, blink_task)
+    await asyncio.gather(button_task, animation_task, blink_task)
```
dhalbert, line 94: Still called `animation_controls` here, and the class is `AnimationControls`. Do you want to change both of those to just `controls`/`Controls`?

kattni: @dhalbert Nope! I want the `AnimationControls` bit to be separate. Because it contains the bits to control the animation. So I made the change to the animation functions only to separate out that they are expecting to control them, and that thing will eventually be `AnimationControls`.

dhalbert (approved):

PR #2139, opened 2022-04-09
Update ANO_Rotary_Encoder_NeoPixel_Arduino_Example.ino

ladyada:
for other platforms


caternuson (approved):

caternuson:
Did not test. But code change looks good, and more in line with the "recommended" way:
https://www.arduino.cc/reference/en/language/functions/external-interrupts/attachinterrupt/

PR #2149, opened 2022-04-22
Ignored hidden filenames created by macOS.

hvtuananh:
  of the code were modified.**  

This change limited the scope of MacroPad Hotkeys to not look at the hidden filenames created by macOS (e.g. macOS will create a file named `._mac-safari.py` for a file named `mac-safari.py`

- **Describe any known limitations with your change.** 
I only tested this on macOS but this should not do any harm on any other platforms.

- **Please run any tests or examples that can exercise your modified code.** 

This works well on my mac. Without this change the serial console will display error messages about malformed macro format for these `._` files.

### DIFF
--- a/Macropad_Hotkeys/code.py
+++ b/Macropad_Hotkeys/code.py
@@ -81,7 +81,7 @@ apps = []
 files = os.listdir(MACRO_FOLDER)
 files.sort()
 for filename in files:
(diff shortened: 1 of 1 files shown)


ladyada (approved):
thank you for the simple and clear PR!

PR #2150, opened 2022-04-25
Adding code for midi friends learn guide

BlitzCityDIY:
Adding code for the USB to Serial midi friends learn guide. Using a pylint ignore for unused-import because the midi libraries are used to recognize the incoming midi messages over USB to send over UART


TheKitty:
You'll need to put in the SPDX lines

TheKitty (approved):

PR #2163, opened 2022-05-16
Use ESP32SPI directly instead of NTP

tannewt:
The NTP library is changing to support native sockets and this API will break.

Needed before https://github.com/adafruit/Adafruit_CircuitPython_NTP/pull/20


ladyada (approved):
yay fyi @brentru

PR #2166, opened 2022-05-24
added NTSC SMPTE composite video code

jedgarpark:
This contains the `smpte_color_bars.ino` sketch inside of the necessary ESP_8_BIT_composite library folder (unchanged from the original so I didn't think it needed forking?). I'm probably doing this wrong from a github standpoint, but should at least get you going @ladyada on making your test pattern nubshank.


jedgarpark:
This is the library required https://github.com/Roger-random/ESP_8_BIT_composite

TheKitty (approved):

TheKitty:
@jedgarpark Merged - please discuss with Ladyada on the license type - we might be able to have our code MIT which links to their GPL lib.

PR #2168, opened 2022-05-25
change numpad example to emulate keyboard numpad

bleything:
The current numpad example emulates a phone keypad. This PR changes it to emulate the numpad section of a keyboard instead, including using the proper keycodes for a numpad and changing the layout to have a double-width `0` key and a period (instead of an asterisk and pound sign).

This allows the macropad to be useful for software that requires a numpad, like with Blender or certain older games.

(it also changes the colors because I'm not a big fan of yellow 😅 )

### DIFF
--- a/Macropad_Hotkeys/macros/numpad.py
+++ b/Macropad_Hotkeys/macros/numpad.py
@@ -11,21 +11,21 @@ app = {                # REQUIRED dict, must be named 'app'
     'macros' : [       # List of button macros...
         # COLOR    LABEL    KEY SEQUENCE
         # 1st row ----------
-        (0x202000, '7', ['7']),
-        (0x202000, '8', ['8']),
-        (0x202000, '9', ['9']),
+        (0x200020, '7', [Keycode.KEYPAD_SEVEN]),
+        (0x200020, '8', [Keycode.KEYPAD_EIGHT]),
+        (0x200020, '9', [Keycode.KEYPAD_NINE]),
         # 2nd row ----------
-        (0x202000, '4', ['4']),
-        (0x202000, '5', ['5']),
-        (0x202000, '6', ['6']),
+        (0x200020, '4', [Keycode.KEYPAD_FOUR]),
+        (0x200020, '5', [Keycode.KEYPAD_FIVE]),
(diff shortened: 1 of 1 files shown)


Macropad_Hotkeys/macros/numpad.py
```diff
@@ -11,21 +11,21 @@
...
-        (0x202000, '3', ['3']),
+        (0x200020, '1', [Keycode.KEYPAD_ONE]),
+        (0x200020, '2', [Keycode.KEYPAD_TWO]),
+        (0x200020, '3', [Keycode.KEYPAD_THREE]),
         # 4th row ----------
-        (0x101010, '*', ['*']),
-        (0x800000, '0', ['0']),
-        (0x101010, '#', ['#']),
+        (0x200010, '0', [Keycode.KEYPAD_ZERO]),
+        (0x200010, '', [Keycode.KEYPAD_ZERO]),
```
dhalbert, line 27: If there's a real double-width keycap installed, then maybe both keys will get pressed, and then two `0`s will get sent. So maybe have line 27 send nothing?

dhalbert (changes requested):
How about not replacing `numpad.py` completely, but instead maybe renaming `numpad.py` to `telephone_pad.py`, change the displayed `'name'`, and add yours as the new `numpad.py`?

bleything:
> How about not replacing `numpad.py` completely, but instead maybe renaming `numpad.py` to `telephone_pad.py`, change the displayed `'name'`, and add yours as the new `numpad.py`?

Sure, no problem! I may not be able to get to that today but I'll update this as soon as possible.

Macropad_Hotkeys/macros/numpad.py
bleything: Ah, that's a great point I hadn't considered. Let me think about that a bit.

dhalbert:
Also, the current numpad rows are not really telephone: they are in this order:
```
789
456
123
*0#
```
but a regular telephone keypad is
```
123
456
789
*0#
```
so that would be something else to change when making `telephone_pad.py`

TheKitty:
Is this still unsettled? I'd like to close this. Thanks!

PR #2169, opened 2022-05-27
qt py esp32 selftest

ladyada:


kattni (approved):
Thanks for adding this!

PR #2176, opened 2022-06-02
Fix consistent connection failures

pjhatesgithub:
The connection to www.adafruit.com:443 is consistently failing. (line 91) Adding
`client.setInsecure()`
right before the connection attempt causes it to always succeed.

I notice all the certificate commands are commented out just above the connection attempt. 
These can't simply be uncommented with the code as it is.


pjhatesgithub:
How the heck did I end up with 78 commits in here? I hate github.

ladyada:
username checks out

PR #2177, opened 2022-06-02
Fix for consistent server connection failures

pjhatesgithub:
The connection to the server consistently fails. Adding
`  client.setInsecure();'
before the request causes it to always succeed.

Above this code some certificate methods are commented out. They can't just be uncommented with the code as is.


ladyada:
this isn't for nina-fw, this is arduino on esp32-s2 - its a recent change. im not actually sure why it doesn't check the certificate chain but its expecting us to store the individual cert fingerprint for each tls connection url (?)
that said, this is OK :)

ladyada (approved):

ladyada:
thanks @pjhatesgithub !

PR #2190, opened 2022-06-15
default neotrellis addresses

jedgarpark:
changed the addresses for I2C to the "default" from the multitrellis guide photo.


TheKitty:
Restarted two failed builds that appear to be an arduino.com file permissions issue, not related to PR

PR #2210, opened 2022-07-02
Reset state on CTRL+C exit.

kattni:
Existing code meant, on CTRL+C exit, if on, the light would remain on, and the buzzer would continue beeping. Updated code to catch the `KeyboardInterrupt`, run the `reset_state()` function, and print the exiting status to the serial console.


FoamyGuy (approved):
Looks good to me. Thanks for the improvement!

PR #2212, opened 2022-07-05
Added completed - cancelled code.

kattni:
Realised while testing that there was no contingency for cancelled Actions runs, and in fact the status light for queued/in progress would continue after the run was cancelled. Chose red blink for the light status for cancelled runs.


TheKitty (approved):
LGTM

PR #2232, opened 2022-07-30
Update MagTag Weather code

caternuson:
Updates MagTag Weather project:
https://learn.adafruit.com/magtag-weather
to improve location specification. Can now specify as either a string location or as a lat/lon tuple for cases where city name is ambiguous:
https://forums.adafruit.com/viewtopic.php?f=60&t=192681

Guide will also need some updating to discuss this new feature.


TheKitty (approved):
lgtm

PR #2233, opened 2022-07-30
changed touch pin so it'll work on Feather RP2040 (was on A4, now on A3)

jedgarpark:
demo code for motorized slide pot on Feather, now on a pin for touch that works (exists) on Feather RP2040.


dhalbert (approved):
lgtm!

PR #2240, opened 2022-08-10
use Pin, not PWMOut, for PulseOut

dhalbert:
When creating a `PulseOut`, CircuitPython 7.x.x encourages a `Pin` instead of a `PWMOut`, and CircuitPython 8 requires it. Fix code that uses the old style.

- @BlitzCityDIY see Owl_IR_TV_Remote changes. 
- @jedgarpark TVBGone also has changes. Also the guide doesn't work on Gemma M0 any more, because we dropped PulseOut from the tiny M0 builds, so I deleted the Gemma-M0-specific example. I will put a warning on the [Gemma M0 page,](https://learn.adafruit.com/circuitpython-tv-zapper-with-circuit-playground-express/gemma-m0-variant) but perhaps we just want to remove it. Also there are a number of feedback items.


TheKitty (approved):

BlitzCityDIY (approved):
Noted, thanks @dhalbert ! Just did a quick test and seems good.

tannewt:
Thanks for doing this!

PR #2246, opened 2022-08-23
Examples for new esp32_camera module in CircuitPython 8

jepler:
These need more polish and the guide is yet to be started, so it's created in draft mode.


TheKitty:
Some import issues

TheKitty (commented):
@jepler - is there a reason the license type is "Unlicense" rather than "MIT"?

jepler:
> @jepler - is there a reason the license type is "Unlicense" rather than "MIT"?

not sure why. I started with a copy of a copy of a file with the copyright style in the boilerplate I used. There are about 44 "unlicense"'d files in learn currently, but I can change it to MIT if that's the standard.

jepler:
Especially as RedHat has decided that CC0 also called the Unlicense is not suitable for "open source" computer code [see https://hackaday.com/2022/07/29/why-fedora-decided-to-give-cc0-licensed-code-the-boot/ and other discussions] I think it's good to fix this and maybe check the authorship on the other 4 dozen files to see if adafruit can change them to MIT at this date.

jepler:
Though, unlicense is not CC0, they are distinct, hm

TheKitty:
Our Adafruit standard is MIT unless otherwise indicated

jepler:
Licenses updated, ready for review

TheKitty (approved):
LGTM

PR #2248, opened 2022-08-23
Updated internal LED code to work better with bundler

evaherrada:


TheKitty (approved):

kattni (changes requested):
Thanks for this, @evaherrada! One change: please remove the comments above the new imports. Since it does things automatically now, there's no need to list the boards, and it's not a full list anyway.

PR #2251, opened 2022-08-24
Switched to mp3s that we own the rights to

evaherrada:
Haven't tested but I can if you want me to


kattni (approved):
@evaherrada Please test just to be certain. Approving anyway.

evaherrada:
Will do

PR #2261, opened 2022-09-02
ez make oven feature enhancement - dynamic profile selection in memory

daneallen:
# Background
I wanted an easier way to switch between the various solder reflow profiles already available on the PyPortal Device that did not require a connection to my computer and a modification of the config.json file. For instance, after running a Sn63/Pb37 profile I might want to run a custom profile for smoothing out Hot Glue that would be used to diffuse LED light. I wanted to be able to make the changes while only interacting with the PyPortal and without making a write change to the on disk config.json file. To that end this change is submitted that supports that effort.

## Note
Being able to dynamically select a profile was limited to being in memory since the code that is running only recognizes the disk as read only. To support dynamic profile selection it has to be done all in memory. I of course discovered this little fact hours into an initial idea that had to be completely refactored to overcome this challenge.

### Changes made are limited to the code.py file and are roughly as follows:
- import os to support collecting the available profiles within the profiles folder
- encapsulate the default_view of the main GUI into a callable function that can yield code execution to the change_profile function and then return to.
    - This required multiple variables to be declared as global as they were used in various other functions. These variables are now grouped together in their declaration and identified at the beginning of requisite function definitions.
- The button for "Profile Change" was placed on unused screen space AND is only present when the oven is in the "Ready" state so as to NOT allow profile changes while actively being run.
- While making changes it was observed that the button.group method had been deprecated and so I removed the use of that method call.
- While reading documentation I noticed a button.contains method that I considered more elegant to use for determining if a given button was pressed. I considered that usage of the button.contains() method would be preferred to mitigate any future issues whereby a button location may be moved without knowing that hard coded button location comparisons where happening elsewhere in code, which would manifest as unresponsive button presses.

### Limitations
- Only 6 profiles are selectable from the change_profile layout. This was deliberate to honor the size constraints of the screen and not make it overly complex with multiple tabs needed for profile selection. More that 6 profiles can be present on the PyPortal, but only the first 6 will be present for dynamic selection.

### Testing
- I ran the code changes multiple times on my assembly that was built identical to [Adafruit's EZ Make Oven](https://learn.adafruit.com/ez-make-oven) instructions. Furthermore I updated my EZ Make Oven codebase to the latest available Adafruit CircuitPython 7.3.3 for testing to ensure that all changes remained functional.

I've used and loved my EZ Make Oven for years and I know this change will enable me to more easily use my Oven. I'm hopeful that others will find the contribution valuable as well. Thank you Adafruit for all that you inspire me to do.

### DIFF
--- a/PyPortal_EZ_Make_Oven/code.py
+++ b/PyPortal_EZ_Make_Oven/code.py
@@ -13,6 +13,7 @@ import audioio
 import audiocore
 import displayio
 import digitalio
+import os
 from adafruit_bitmap_font import bitmap_font
 from adafruit_display_text import bitmap_label as label
 from adafruit_display_shapes.circle import Circle
@@ -21,13 +22,10 @@ import adafruit_touchscreen
 from adafruit_mcp9600 import MCP9600
 
 TITLE = "EZ Make Oven Controller"
-VERSION = "1.3.2"
+VERSION = "1.3.3"
 
 print(TITLE, "version ", VERSION)
 time.sleep(2)
-
-display_group = displayio.Group()
-board.DISPLAY.show(display_group)
 
 PROFILE_SIZE = 2  # plot thickness
 GRID_SIZE = 2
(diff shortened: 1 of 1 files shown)


daneallen:
Come on... this is a pretty cool feature to include.

cogliano (approved):
Sorry for the delay, I finally had some time to look at this in detail. Overall it looks fantastic! I like how the button disappears when the oven is in use. I think I will change the names of the JSON files to be more user friendly but that can come later. Thanks for making this improvement to the EZ Make Oven.

TheKitty (approved):

PR #2267, opened 2022-09-05
Code for new outside guide Tiny LED WiFi cube

TheKitty:

  the code.

  process.


kattni (commented):
Looks good to me, but much of this is over my head. Dan or Jeff might have some insight, if any is needed. If they choose not to take a look, let me know and I'll approve.

QT_Py_Cube/cube.py
```diff
@@ -0,0 +1,185 @@
...
+        now = time.monotonic()
+        if now >= self.last_color_time + self.color_wait:
+            self.color_idx = self.color_idx + 1
+            self.last_color_time = now
+
+        return result
+
+    @staticmethod
+    def __convert_to_hex(color):
+        return int('0x%02x%02x%02x' % color, 16)
```
dhalbert, line 185: A color is an integer, so this `__convert_to_hex()` is not needed: it converts an integer to hex and then back to an integer again. So just use the color name variable (like `CYAN`) and you should not need this routine at all.

dhalbert (changes requested):

QT_Py_Cube/cube.py
TheKitty: Have asked the author to review

chardane: Thanks for the review! I was able to remove a lot of usages of this `__convert_to_hex()` function, but unfortunately still need it in two places -- for some reason, when i pass in just a color (e.g. CYAN) to `PixelFramebuffer.text` or `PixelFramebuffer.rect`, I get the following error:

```
  File "adafruit_framebuf.py", line 318, in rect
  File "adafruit_framebuf.py", line 165, in fill_rect
TypeError: unsupported types for __rshift__: 'tuple', 'int'
```

It's not immediately apparent to me what this error means, but i do seem to get around it by using this convert method. I've sent an updated copy of the cube.py file, happy to discuss if more changes are needed as well, thanks!

dhalbert: I was wrong. The colors are tuples: `(R, G, B)` values. You can use those many places, but for for those locations, you can use `int.from_bytes()`, which will be faster than formatting a string:
https://github.com/todbot/circuitpython-tricks#convert-rgb-tuples-to-int-and-back-again
```py
rgb_tuple = (255, 0, 128)
rgb_int = int.from_bytes(rgb_tuple, byteorder='big')
```

chardane: ooh nice! unfortunately I just tried this:

```
@staticmethod
    def __convert(color):
        return int.from_bytes(color, byteorder='big')
```

and it gave me this error:
```
TypeError: function doesn't take keyword arguments
```

tried with `int.from_bytes(color)`
and got `TypeError: function missing 1 required positional arguments`

and trying it with `int.from_bytes(color, 'big')`
got the error `TypeError: object with buffer protocol required`

I did look at the [python stdtypes documentation](https://docs.python.org/3/library/stdtypes.html) and it does look like what you suggested should have worked, but it also looks like the circuitpython docs say something else:
https://docs.circuitpython.org/en/latest/docs/library/builtins.html?highlight=byteorder#int
> In CircuitPython, byteorder parameter must be positional (this is compatible with CPython).

So I do have to pass _something_ in as a positional parameter, just not sure which object that needs to be?

I also did a search for `int.from_bytes` in this repo, and it looks like there's only two comments referencing this function, to say that the function doesn't support "LE" (and I'm not sure what that means 😅)  

Should I just keep the code as is for now since it seems like it might be the simplest path forward even though it's not the fastest it can be? :D

Screenshot of repo search below:

Screen Shot 2022-09-06 at 10 03 50 PM

chardane: Thanks so much! Just tested it and it worked -- changes are incoming :)

Thanks for the help, and I'm glad I learned something new this week 🥳

TheKitty:
Ok @dhalbert - I have uploaded the changed cube.py file for @chardane - linting now

dhalbert (approved):
Thanks for your patience!

PR #2269, opened 2022-09-06
adjusted for Feather RP2040

jedgarpark:
Updated to use Feather RP2040 (left in a line to uncomment for Feather M4), changed rotary encoder I2C address to default 0x36, added audio.pause()/audio.resume() between track changes to remove noise.


TheKitty (approved):
lgtm

PR #2280, opened 2022-09-16
update UART-Between-Boards guide to work around NRF52 bug before CP8

Neradoc:
On Circuitpython before 8, UART.read() on NRF52 return `b''` instead of `None` if no byte was read.


dhalbert (approved):
Thanks! Simple fix, and won't need to change it back.

Issue #2266, opened 2022-09-03
MagTag weather: Open weather API limitations to new accounts


Neradoc:
Hi,
In the MagTag open weather project, it seems that the onecall APIs (get_forecast) are not accessible to a new free account anymore and require to subscribe to the onecall service, which is "free" but requires registering a card to pay-per-call above the allotted free ones. The default free accounts only include the current weather and 3-hours forecast, not the daily. <https://openweathermap.org/price#weather>

Using a new account, after making sure the API key is active, as it works with the calls to the free Geocoding API, I get the following response from the site:
```py
{'message': 'Invalid API key. Please see http://openweathermap.org/faq#error401 for more info.', 'cod': 401}
```

Which is described as what the API returns when not authorized to access in the mentioned link (other causes being ruled out).
> You are using a Free subscription and try requesting data available in other subscriptions . For example, 16 days/daily forecast API, any historical weather data, Weather maps 2.0, etc). Please, check your subscription in your personal account.

This subsequently leads to:
```py
Traceback (most recent call last):
  File "code.py", line 280, in <module>
  File "code.py", line 92, in get_forecast
KeyError: daily
```

When I use the account I created way back when the guide was new it allows me to access the onecall API and the code runs as expected. I think the guide has to be revised to mention that new limitation, or another API has to be found.

Note that the free tier has "3-hour" forecast over 5 days (`https://api.openweathermap.org/data/2.5/forecast?&lat=&lon=`) so we could loop through that to find the min-max temperatures over a day for example, but the rest would be more complicated.

https://learn.adafruit.com/magtag-weather
https://github.com/adafruit/Adafruit_Learning_System_Guides/tree/main/MagTag_Weather

tlyu:
> This subsequently leads to:
> 
> ```python
> Traceback (most recent call last):
>   File "code.py", line 280, in <module>
>   File "code.py", line 92, in get_forecast
> KeyError: daily
> ```

I think the code should also check the HTTP status code, and give a more useful error here, instead of trying to look up JSON keys as if there were always a successful response.

caternuson:
Was the 2.5 API being used in all cases? I ran into something similar here:
https://forums.adafruit.com/memberlist.php?mode=viewprofile&u=345839

The key I created way back when writing the guide still works with 2.5, but not 3.0. As long as the 2.5 API is used, the guide still works as shown (with that key at least).

[...]

PR #2290, opened 2022-10-03
Update for MagTag Weather - Add new free API code version

caternuson:
This is for #2266 

[Openweathermap](https://openweathermap.org) changed their free API access and the One Call API that was being used is no longer available for free. This PR adds a new version of the code that attempts to use the still free Forecast API. The Forecast API data return is structured differently than the One Call API, so there are some quirks and caveats.

The new code is under the forecast directory.

The old code is retained under the onecall directory. It should still work for anyone that obtains a non-free API key.


TheKitty:
image

TheKitty (approved):

PR #2294, opened 2022-10-04
Override ArduinoHttpClient library

PaintYourDragon:
Addresses an issue where current release of ArduinoHttpClient fails Arduino CI due to syntax. Appears to be fixed in library’s main branch but not yet an updated release…when that occurs, this edit can be removed.


brentru:
Looks good to me and seems to pass CI.

PR #2299, opened 2022-10-08
Update code.py

dwalton64:
Openweather is now using API 3.0 and I had to make this change to the data source URL to get this code to successfully retrieve data.

### DIFF
--- a/MagTag_Weather/onecall/code.py
+++ b/MagTag_Weather/onecall/code.py
@@ -63,7 +63,7 @@ def get_data_source_url(api="onecall", location=None):
         URL += "&lat={}".format(location[0])
         URL += "&lon={}".format(location[1])
     elif api.upper() == "ONECALL":
-        URL = "https://api.openweathermap.org/data/2.5/onecall?exclude=minutely,hourly,alerts"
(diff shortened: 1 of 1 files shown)


caternuson:
What error message did you receive when trying to use the 2.5 API?

dwalton64:
Looking back at the error, I see that it was a key error.  I suspect it
might have been because I had just created my key and set up the onecall
charges.  I changed the URL to use the 3.0 API and it worked immediately so
I figured I could not add a key that would work with the old API.  But it
worked just now when I changed it back to 2.5.

I suspect that my timing was really good and that my key finally got
accepted right at the moment I changed the URL to use the 3.0 API.

In any case, the 3.0 API seems to work fine with the code, but it is not
broken with the 2.5 API as I thought.  I apologize for my confusion.

Andrew

caternuson:
Yep, does sound like just a delay in key activation. No worries. There's been some recent changes in the API access. Some older accounts/keys seem to be able to still access the One Call API. Newer accounts require setting up payments. The guide was recently updated with a new version of the code that uses the Forecast API, which still seems to be available for free.
https://learn.adafruit.com/magtag-weather/project-code#installing-project-code-3079586

PR #2305, opened 2022-10-16
speculative fix for example build failure

jepler:
This example was failing; I think it might be due to a new SdFat version, and that this might be the fix.


TheKitty:
Mini_GIF_Players.ino rp2040:rp2040:adafruit_feather has errors related to SDFat

jepler:
OK, this time it builds. I dunno if @BlitzCityDIY has time to check if it still works or if it's better to merge it anyway.

TheKitty (approved):

PR #2326, opened 2022-11-04
Use Python 3.10 during CI; 3.11 is incompatible with pylint 

jepler:


jepler:
Typical error message:
```
ImportError: cannot import name 'formatargspec' from 'inspect' (/opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/inspect.py)
Error: Process completed with exit code 123.
```

TheKitty (approved):

PR #2328, opened 2022-11-04
Use a different concurrency group name

### LINKED ISSUE #2327: change workflow concurrency group names to avoid GA weirdness
I received an update from GH support about our problem where the image updater is cancelled ...I think we can change our workflow files so that this doesn't happen.

> Thank you for your patience while the team investigated this.
> 
> I have been updated that the workflow runs are actually getting canceled by pages-build-deployment runs. For example, https://github.com/adafruit/Adafruit_Learning_System_Guides/actions/runs/3376580648 was canceled by https://github.com/adafruit/Adafruit_Learning_System_Guides/actions/runs/3376591433.
> 
> pages-build-deployment actually also uses concurrency group feature under the cover to prevent multiple pages build from running. It currently uses the source branch name as group name.
> 
> This scheduled workflow has a job that [pushes to the branch folder-images](https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/e85845c776dd7190b8c071453dece6f69562eade/.github/workflows/images.yml#L52). This push will trigger pages-build-deploymentworkflow. Pages workflow will start running also using folder-images as concurrency group name because it’s on folder-images branch. If pages workflow started before the scheduled workflow fully completes, it will attempt to cancel the scheduled workflow causing the scheduled workflow to report cancelled status.
> 
> The Pages team will work on changing the concurrency group name they use in pages-build-deployment to something less generic to avoid collisions like this in the future.
> 
> As a workaround in the meantime, could you use a different concurrency group name in your scheduled workflow to mitigate this issue?
> 
> I hope this helps, please let us know if you need any further assistance.

jepler:
.. the other one is haunted.  Closes #2327.

Also upgrades actions/checkout to one that doesn't warn
```
Warning: The `save-state` command is deprecated and will be disabled soon.
Please upgrade to using Environment Files.
For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

### DIFF
--- a/.github/workflows/githubci.yml
+++ b/.github/workflows/githubci.yml
@@ -56,7 +56,7 @@ jobs:
 
 
     - name: Upload build artifacts
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v3
       with:
         name: ${{ github.event.repository.name }}.${{ github.sha }}
         path: |
--- a/.github/workflows/githubci.yml
+++ b/.github/workflows/githubci.yml
@@ -10,7 +10,7 @@ jobs:
       with:
         python-version: "3.x"
     - name: Checkout Current Repo
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
 
     - name: check SPDX licensing
       run: python ./SPDX.py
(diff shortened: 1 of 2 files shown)


FoamyGuy (approved):
Looks good to me and makes sense based on their feedback that was pasted in the linked issue.

Thanks for reaching out to them to look into it. It's nice to see they found and shared such thorough details.

jepler:
> Looks good to me and makes sense based on their feedback that was pasted in the linked issue.
> 
> Thanks for reaching out to them to look into it. It's nice to see they found and shared such thorough details.

agreed! It took a bit of time but github got us an actual answer.

PR #2337, opened 2022-11-22
Add NeXT to USB adapter with CircuitPython & RP2040

jepler:
In 2012, Lady Ada wrote a Learn project to convert a NeXT keyboard to USB with an Arduino Micro.

Now, ten years later, the keyboard is still as awesome as ever!  So awesome that for the anniversary we re-did the project with our favorite environment (CircuitPython) and an awesome microcontroller (RP2040)

In draft state for the moment, as I'm going to move the project from the Adafruit Feather RP2040 to the Adafruit QT Py RP2040 instead! It also doesn't handle sending vertical bar or backslash keys as those are in non-standard shifted locations.

This pull request text was typed on the NeXT keyboard!

### DIFF
--- a/CircuitPython_NeXT_Keyboard_RP2040/code.py
+++ b/CircuitPython_NeXT_Keyboard_RP2040/code.py
@@ -5,12 +5,18 @@ import board
 import rp2pio
 import usb_hid
 from adafruit_hid.consumer_control import ConsumerControl
-from adafruit_hid.consumer_control_code import ConsumerControlCode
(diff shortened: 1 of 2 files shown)


TheKitty:
Please add SPDX

TheKitty:
Per Slack, waiting until review requested

jepler:
@TheKitty Ready for your review!

kattni (approved):
Looks good to me!

PR #2339, opened 2022-11-28
fix httpserver imports. print errors if they occur

FoamyGuy:
This PR: https://github.com/adafruit/Adafruit_CircuitPython_HTTPServer/pull/25 changed the way imports for this library need to be coded. This PR changes to the new import syntax for the only current Learn project that uses the library. 

I've also added a print statement to print errors if they are caught so that they won't be silently consumed. This can make it easier to troubleshoot if something has gone wrong.

### DIFF
--- a/PicoW_CircuitPython_HTTP_Server/code.py
+++ b/PicoW_CircuitPython_HTTP_Server/code.py
@@ -16,7 +16,8 @@ from adafruit_display_text import label
 import adafruit_displayio_ssd1306
 import adafruit_imageload
 from digitalio import DigitalInOut, Direction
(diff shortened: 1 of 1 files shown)


FoamyGuy:
@BlitzCityDIY if you get a chance can you try this out? The httpserver library was updated after this project was published and the code needs tweaked slightly to work with the new one. I tested a minimal version, but I don't have the full project set up to try it out in the real environment.

BlitzCityDIY:
hihi @FoamyGuy - thanks so much! just tested and everything is working great.

BlitzCityDIY (approved):

PR #2347, opened 2022-12-08
Add EYESPI graphics example for Arduino

PaintYourDragon:
Two different display drivers (ST7789, ILI9341) are currently used by EYESPI products, and perhaps more in the future. Rather than add same example to both display libraries (which would inevitably diverge), this creates a single example for _all_ color EYESPI products, selectable with a product ID #define at the top of the code. The naming of directories is such that other display types (mono OLED, eink, etc.) could get their own folders if those products get EYESPIified later.


TheKitty:
Please add SPDX info https://learn.adafruit.com/submitting-arduino-code-to-the-adafruit-learning-system?view=all#add-author-and-license-information

PaintYourDragon:
Fixed now, thx!

TheKitty (approved):

PR #2352, opened 2022-12-09
Update to functional URL

axelmagnus:
Well, they have changed it to https..


BlitzCityDIY (commented):
hihi- thanks for the PR! I tested and both `http` and `https` are working, is there an issue you ran into as to why you want to change it to `https`?

axelmagnus:
Hmm. i got 410 Not  authorized  for the http.

axelmagnus:
latest CPY 7.x btw

PR #2354, opened 2022-12-13
Adding code examples for MOSFET driver

BlitzCityDIY:
Adding simple blink style examples for MOSFET driver

### DIFF
--- a/MOSFET_Driver_Examples/CircuitPython/code.py
+++ b/MOSFET_Driver_Examples/CircuitPython/code.py
@@ -1,0 +1,19 @@
+# SPDX-FileCopyrightText: 2022 Liz Clark for Adafruit Industries
+#
+# SPDX-License-Identifier: MIT
+
+import time
+import board
+from digitalio import DigitalInOut, Direction
+
+# motor output
+solenoid = DigitalInOut(board.D5)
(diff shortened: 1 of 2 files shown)


BlitzCityDIY:
Seems to not like `Serial`, should i add skip files for the affected boards? although shouldn't those boards be okay with `Serial`?

TheKitty:
Hey @caternuson - do you see why it dislikes Serial on some boards and not others?

caternuson:
Looks like a combination of things.

* For the nRF boards (CPB included) and ones with `_tinyusb` an additional include is needed ([see here](https://learn.adafruit.com/introducing-the-adafruit-nrf52840-feather/arduino-board-setup#4-run-a-test-sketch-2860620))
* For gemma and trinket (old versions of both) - they don't have serial
* Not sure why the `cpx_ada` build failed. Possibly just a server hiccup.

I wouldn't suggest adding the nRF specific include to the sketch. Two options (pick one):
* Add `.test.skip` files for each of the failing targets
* Add `.test.only` for good general target(s). Sketch looks very simple. I think OK to just rely on an UNO test, which would be `uno.test.only`

TheKitty (approved):

PR #2372, opened 2022-12-31
Make the CPX Simple Simon compatible with CPB

Neradoc:
This makes the code for the Circuit Playground Simple Simon compatible with the CPB.
I don't think there is another version of it that already works ?
https://learn.adafruit.com/circuit-playground-simple-simon/code-walk-through

- use `cp` not `cpx` from the `adafruit_circuitplayground` library
- do not read the analog pin A7 (not available on CPB)
- reference `cp.touch_TX` (available on both) instead of `cp.touch_A7`

Also:
- fix the call to `game_lost()` which takes a step and not a sequence[setp]
- use a context manager to setup and deinit the analog pins
- some spaces at the end of lines

### DIFF
--- a/CPX_Simple_Simon/code.py
+++ b/CPX_Simple_Simon/code.py
@@ -14,7 +14,7 @@ import random
 import math
 import board
 from analogio import AnalogIn
-from adafruit_circuitplayground.express import cpx
+from adafruit_circuitplayground import cp
 
 FAILURE_TONE        = 100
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Looks good, thanks! Does the guide text need updating?

dhalbert:
> Does the guide text need updating?

I updated the guide text: just needed to change the alert that said it only works on CPX.

PR #2373, opened 2023-01-02
Fix for compatibility with arduinojson 6200

jepler:
I noticed a failure during the overnight build.  It seems that `is<char *>` was deprecated and now removed. `is<const char *>` seems to be the new way. Only compile-tested. See https://arduinojson.org/v6/api/jsonvariant/is/


dhalbert:
@PaintYourDragon -- asking for your review mostly to alert you to this.

TheKitty (approved):

TheKitty:
shoot, sorry Dan

PaintYourDragon:
Noted. Thank you!

PR #2379, opened 2023-01-11
adding demo code for PCA9546

BlitzCityDIY:
Adding demo code since it will be in two guides (breakout and stemma breakout)


TheKitty:
Yes @BlitzCityDIY I believe they can

TheKitty (approved):

PR #2392, opened 2023-01-30
Add code for the sciorpio holiday tree project

jepler:
@djecken Please verify the code still works as expected. I modified it slightly based on pylint messages.


djecken (approved):
Updated code reviewed and works as expected on the build. Thank you!

jepler:
Merging even though there's a problem building some other example.  See https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/2393 which can't totally fix the problem yet.

PR #2393, opened 2023-01-30
BSEC_Software_Library version 1.7.x is incompatible & broken

jepler:
.. in addition, the library release is badly broken (can't find its own include files in the arduino environment).

Once they issue an updated release that's not just totally broken, though, we'll have to adapt to their incompatible rename changes.

I don't have a device to test with but this at least builds, if you move/rename header files in BSEC_Software_Library. But for now the CI will fail due to the brokenness of the library.

### DIFF
--- a/BME680_BSEC_OLED_Demo/BME680_BSEC_OLED_Demo.ino
+++ b/BME680_BSEC_OLED_Demo/BME680_BSEC_OLED_Demo.ino
@@ -50,7 +50,7 @@ void setup() {
   display.setTextColor(SSD1306_WHITE);
   display.setRotation(0);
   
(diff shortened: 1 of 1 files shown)


jepler:
Someone's "reported" this to bosch and the response so far is a non-response: https://community.bosch-sensortec.com/t5/MEMS-sensors-forum/bsec-datatypes-h-not-found-in-arduino-example-for-BSEC-Library/td-p/69438

BSEC2 seems to be  https://github.com/boschsensortec/Bosch-BSEC2-Library -- dunno whether the guide code should be rewritten for this, but it's a possibility to consider since it's already broken.

PR #2395, opened 2023-02-01
updating white noise rocket

BlitzCityDIY:
adding pin for fan output


TheKitty (approved):

BlitzCityDIY:
@TheKitty should i add a trinket_m0 skip file because of the issue with that library?

PR #2403, opened 2023-02-09
Use pylint-learn as the pylint version, if installed

jepler:
.. this allows a user to install the specific version of pylint required, under the specific name:
```
pipx install --suffix=-learn pylint==2.7.1
```
and run pylint_check more easily locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,3 +20,14 @@ This is a work in progress.
 
 [learn]: https://learn.adafruit.com/
 
+## Running pylint locally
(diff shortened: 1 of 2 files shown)


TheKitty:
In README.md, pip instal pipx only has one l

TheKitty:
README still needs a fix on instal->install

jepler:
Fixed @TheKitty !

TheKitty (approved):

PR #2404, opened 2023-02-09
Removed redundant secrets reference for wifi communication

TrevKnows:
I needed to remove this redundant code for the boxing glove learn guide.

### DIFF
--- a/CPBoxing_Glove_Tracker/code.py
+++ b/CPBoxing_Glove_Tracker/code.py
@@ -13,11 +13,10 @@ from adafruit_io.adafruit_io import IO_MQTT
 from adafruit_adxl34x import ADXL345
 from adafruit_lc709203f import LC709203F, PackSize
 
-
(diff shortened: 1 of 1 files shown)


TheKitty:
image

BlitzCityDIY (changes requested):
hihi @TrevKnows we should use **settings.toml** instead of **secrets.py**, so please remove the `try/except` for `secrets` and then:

- add `import os`

- lines 22-23 replace with: 
```
aio_username = os.getenv('aio_username')
aio_key = os.getenv('aio_key')
```

- and then in `MQTT()` update the SSID and password import with 

```
username = os.getenv('CIRCUITPY_WIFI_SSID')
password = os.getenv('CIRCUITPY_WIFI_PASSWORD')
```

TrevKnows:
@BlitzCityDIY on it

PR #2406, opened 2023-02-10
Fix pylint errors that cropped up

TheKitty:


BlitzCityDIY (approved):
was just about to do this, thanks @TheKitty !

PR #2417, opened 2023-02-15
Adding delay

BlitzCityDIY:
Adding delay for debouncing.

### DIFF
--- a/NeoKey_BFF_Examples/CP_Keyboard_Example/code.py
+++ b/NeoKey_BFF_Examples/CP_Keyboard_Example/code.py
@@ -44,8 +44,9 @@ while True:
 
     # if the button is pressed..
     if not switch.value and not switch_state:
-        # neopixel brightness is 0.3 and rainbow animation is visible
         pixels.fill(pixel_color)
         pixels.show()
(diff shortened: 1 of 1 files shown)


NeoKey_BFF_Examples/CP_Keyboard_Example/code.py
```diff
@@ -44,8 +44,9 @@
 
     # if the button is pressed..
     if not switch.value and not switch_state:
-        # neopixel brightness is 0.3 and rainbow animation is visible
         pixels.fill(pixel_color)
         pixels.show()
         keyboard.press(modifier_key, key)
         switch_state = True
+        
+    time.sleep(0.05)
```
dhalbert, line 52: Add an `elif` to line 46 otherwise the `time.sleep()` only affects the first test but not the second. e.g. there could be a bounce between lines 43 and 46.

dhalbert (changes requested):
Thanks, one change.

TheKitty (approved):

TheKitty (approved):

dhalbert (approved):
Thanks!

PR #2458, opened 2023-03-27
Adding MCP2515

BlitzCityDIY:
Add new Adafruit MCP2515 library to dependencies


BlitzCityDIY:
Library is not published, can't add yet as dependency

PR #2459, opened 2023-03-28
improve guide code

jepler:
* Use the LED built into the button
 * blink while getting text, solid when you can press to get another one
 * clear out extra button presses
 * allow auto restarting after most exceptions


jepler:
fixed 3 pylint problems, hopefully that was all

TheKitty (approved):

PR #2471, opened 2023-04-05
Add additional files for Learn Guide

TheKitty:


jepler (approved):
It's better if each example is is own directory/folder with a code.py file,but this structure is acceptable and should work okay with bundlefly.

PR #2497, opened 2023-05-16
Make Arduino build conditional and scheduled

tekktrik:
@kattni - I mocked changes for what I think might be a good shift.  Happy to modify, but I wanted to have something to point to.

This breaks out the Arduino build steps to their own file, which now only runs when at least one of these is true:

* It is 3 am (daily)
* A PR is submitted that updates a file with extensions `.ino`, `.c`, `.cpp`, `.h`, or `.hpp`
* The workflow is manually triggered to run

### DIFF
--- a/.github/workflows/arduino_cron.yml
+++ b/.github/workflows/arduino_cron.yml
@@ -1,0 +1,128 @@
+name: Arduino Library CI
+
+on:
+  schedule:
+    - cron: 0 3 * * *
+  workflow_dispatch:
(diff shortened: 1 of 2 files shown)


tekktrik:
Okay, this is ready!  As discussed I can start a new work package for the platform+project specific builds for Arduino when submitted through PR after this one.

kattni (approved):
This is AMAZING. Thank you so much!

PR #2503, opened 2023-05-17
Add code for rpi pirate radio

caternuson:
Code and assets for Raspberry Pi based "pirate" radio guide.


ladyada (approved):
do you wanna try adding RDS support?

caternuson:
Added to the extent supported by current library. But can only get station name to show up on my test receiver. Same thing I saw before, from what I remember, exactly what the "rds buffer" is not well documented.

PR #2511, opened 2023-05-22
Changes for httpserver 4.0

FoamyGuy:
These changes are required for this update https://github.com/adafruit/Adafruit_CircuitPython_HTTPServer/pull/54 in the HTTPServer library. 

This is the only project currently that uses the library that I could find by searching the Learn repo and the Library Bundle Examples.

I do not have all of the exact same hardware to test the full project. I did test a version that was modified to to use a built-in display and to use hardcoded temperatures instead of reading from a sensor. With this version I was able to verify the webserver  portions of the code are working as expected with these changes. 

I think it should be working on the real hardware, but it may be good to try that out to confirm if possible.

### DIFF
--- a/PicoW_CircuitPython_HTTP_Server/code.py
+++ b/PicoW_CircuitPython_HTTP_Server/code.py
@@ -16,11 +16,7 @@ from adafruit_display_text import label
 import adafruit_displayio_ssd1306
 import adafruit_imageload
 from digitalio import DigitalInOut, Direction
-from adafruit_httpserver.server import HTTPServer
-from adafruit_httpserver.request import HTTPRequest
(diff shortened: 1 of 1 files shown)


FoamyGuy:
There are 2 or 3 code sections that are in the guide on this page: https://learn.adafruit.com/pico-w-http-server-with-circuitpython/code-the-pico-w-http-server that will need updated at the time that this gets merged. 

The top portion of the page is an embed of the repo file so it will update when this gets merged. But the sections underneath that include the explanations have code that is held directly in the guide rather than the repo so it'll need updated seperately.  

I can make the changes in the guide when the time comes if it is helpful.

FoamyGuy:
Closing this one, resolved by: #2527

PR #2529, opened 2023-06-12
Fix key release pixel effect.

kattni:
Before it meant that a key release would turn off both LEDs, even if the other key was still pressed. Now the release event only turns off the LED associated with that key.


TheKitty (approved):

TheKitty (commented):
LGTM

PR #2543, opened 2023-06-23
Update BME680 for latest BSEC release

jepler:
compile-tested with 1.8.1492. Not tested on hardware.

Besides testing on HW, the related guide page needs to be updated, as it calls for the prior version 1.6.1480 to be installed instead: https://learn.adafruit.com/adafruit-bme680-humidity-temperature-barometic-pressure-voc-gas/bsec-air-quality-library

besides the pinned version I think that this also needs to change
> Change BME680_I2C_ADDR_PRIMARY to BME680_I2C_ADDR_SECONDARY

these are now called ADDR_HIGH and ADDR_LOW in the library


TheKitty (approved):

caternuson:
HW tested with a QT PY M0, PID 3660 BME680 breakout, and PID 938 OLED using the 1.8.1492 release of the BSEC library. It works, output on OLED, etc.

TheKitty (approved):
LGTM

PR #2557, opened 2023-07-10
Fix crash on days without a moon rise or set

PaintYourDragon:
Rare occurrence (maybe once/month) but it happens. sandorcourane spotted this and provided the fix in the forums.


dhalbert (approved):
Thanks. This looks OK, assuming that `moonrise` and `moonset` are always present. The previous code did not assume they were.

PR #2565, opened 2023-07-13
Add print to identify monitor type.

kattni:
Tested successfully on Feather ESP32-S2.


KeithTheEE (approved):
Looks good to me and makes the code more explicit about the battery monitor.

PR #2571, opened 2023-07-19
[Small bugfix] ESP32 WiFi example

elpekenin:
- Fixed the code to properly handle timeout, as `ping()` returns `None` under that scenario and code throws an exception trying to multiply it by 1000 (converting s to ms). 

    Also added a couple comments for further understanding of the code.

- This should preserve the logic,except for:
    - Not running into exceptions
    - Explicit message when (if) we timeout twice, could maybe add a loop for more retries...


dhalbert (approved):
Thanks!

PR #2582, opened 2023-07-31
compile ESP32 passthrough utility with TinyUSB

dhalbert:
NINA-FW 1.7.5 failed to upload completely on MatrixPortal M4, PyPortal M4, etc. See https://github.com/adafruit/nina-fw/issues/53. After experimentation, I narrowed down the problem to something involving the Adafruit SAMD Arduino USB stack. Using TinyUSB instead fixes the problem. New versions of the passthrough were uploaded to https://learn.adafruit.com/upgrading-esp32-firmware/upgrade-all-in-one-esp32-airlift-firmware#upload-passthrough-code-3124109.


dhalbert:
Hmm, I have to set the USB Stack option for the build:
```
2023-07-31T00:37:57.1163264Z /home/runner/Arduino/libraries/Adafruit_TinyUSB_Library/src/Adafruit_TinyUSB.h:32:2: error: #error TinyUSB is not selected, please select it in "Tools->Menu->USB Stack"
```
Not sure how to do that in the CI.

brentru:
@dhalbert  You'd want to target the platforms with `_tinyusb` attached such as `pyportal_tinyusb` (https://github.com/adafruit/ci-arduino/blob/master/build_platform.py#L116). 

I'm unsure how we'd do this within the Learning System Repo as we test specific platforms (https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/.github/workflows/arduino_cron.yml#L63). Possibly adding a .skip file would be OK here unless you want artifacts.

tyeth:
Yeah, using the `.<platform target>.test.only` files per target (from ci-arduino/all_platforms.py) and a .target.generate file per uf2.
I've just merged in the tinyusb M4 airlift target to the ci matrix for this repo so that should be available now.
https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/3286/changes

PR #2583, opened 2023-08-01
Make a QR code demo for pycamera

jepler:
this requires the (unreleased) PyCamera module, so, creating as a draft PR.


TheKitty (approved):

jepler:
I moved this into the PyCamera repo instead, so it does not need to be added to learn at this time.

PR #2588, opened 2023-08-11
Adding arduino Metro RP2040 SD card example

BlitzCityDIY:
Adding Arduino SD card example for Metro RP2040


BlitzCityDIY:
converting to draft to revisit getting it to pass CI monday. may need to switch libraries.

ladyada:
please use adafruit SDfat fork library
https://github.com/adafruit/SdFat

PR #2593, opened 2023-08-17
remove unneeded workaround

jepler:
The underlying issue was resolved:
 * https://github.com/adafruit/Adafruit_CircuitPython_Display_Text/issues/183


TheKitty:
FYI @jepler

PR #2594, opened 2023-08-18
2581 python packaging

anotherthomas:
The installation of the magic ai storybook software is quite complex and manual. 
Packaging the software allows to 
a) automatically download dependencies and 
b) create entry points to start the software like a regular program.

Installing pyaudio via pip requires portaudio19-dev, otherwise you can install the debian package


makermelissa:
Shortly after you made this, bookworm was released and the code no longer worked. I have just updated to run on bookworm, so I can now test this out.

makermelissa:
Ok, tested this today. It seems to install correctly and with a venv active, it installs to the venv bin folder. However, when running it, nothing displays on the screen. I don't know if you are wanting to update/test this yourself. If not, we can leave the guide as is because installing the packages manually isn't much more than copying/pasting the commands at the moment.

makermelissa (changes requested):
Please test/update for bookworm and using a venv. Thanks.

anotherthomas:
I don't have the screen here at the moment, will do so as soon as possible

PR #2605, opened 2023-08-29
Talking D20 projects

PaintYourDragon:
This PR contains files for TWO VERSIONS of the Talking D20 project (as explained in the READMEs in each folder).

For some reason, the original Arduino code was still a separate repo rather than in the Learn guides collection. That repo will be removed once this is in place here and the guide updated to match.

CircuitPython code is new.


TheKitty:
If the library isn't in the build yaml, there is a guide page on adding it 
https://learn.adafruit.com/submitting-arduino-code-to-the-adafruit-learning-system/missing-libraries

TheKitty:
I cannot tell, looks like you need a ".test.only" file for the architecture you are compiling to, it's in that same guide.

TheKitty (approved):

TheKitty:
Thanks Phil, looks good

PR #2606, opened 2023-08-29
Update settings.toml to use cp expected name

makermelissa:
Since the settings.toml in this test is used in guide pages to represent the default setting.toml, it should use the correct names of the settings. See https://docs.circuitpython.org/en/latest/docs/environment.html#environment-variables. @kattni, since you last updated this file, I'll just tag you to ensure it doesn't break something.


kattni (approved):
Yeah it's fine. I should have spoken up about this when we decided on the standard variables for the web workflow, but I don't recall the conversation. As long as they match, it won't break anything.

makermelissa:
Thanks

PR #2611, opened 2023-08-31
Set Fan Motion Outlet code to use MQTT by default

makermelissa:
While updating the code, I realized the guide expects this value should be set to True by default to make it easier for users and somehow it was changed to False along the way (probably while testing something).


TheKitty (approved):
LGTM

PR #2624, opened 2023-09-15
minor bug fix

BlitzCityDIY:
update to add a 0 in front of single digit minute times


caternuson:
could just set width of 2 with zero padding and it'll work for all values:
```python
>>> minutes = 8
>>> f"{minutes:02}"
'08'
>>> minutes = 23
>>> f"{minutes:02}"
'23'
>>> 
```

PR #2643, opened 2023-10-11
Update ESP32 Arduino examples for LEDC usage

caternuson:
Updates several Arduino examples to latest (3.x) ESP32 BSP that has a different API for LEDC. Older code left in place with new code added and surrounded with preproc logic.

All examples tested.


makermelissa (approved):
Thank you.

PR #2644, opened 2023-10-12
Add test skip for failing Qualia build

caternuson:
This will skip the build test for Qualia ESP32 S3 as temporary CI fix until things catch up with latest ESP32 BSP release.


makermelissa (approved):
Thank you

PR #2649, opened 2023-10-17
first commit bricktunes code

jedgarpark:
first commit bricktunes LEGO color synth glove. The `cedargrove_chime` library is in the community bundle, but i think it has a path typo (`liibraries`) at the moment, so not sure if that'll cause any issues. I've let @CedarGroveStudios know about the issue.

### DIFF
--- a/Bricktunes_LEGO_Color_Synth/code.py
+++ b/Bricktunes_LEGO_Color_Synth/code.py
@@ -1,0 +1,142 @@
+# SPDX-FileCopyrightText: 2023 John Park for Adafruit
+# SPDX-License-Identifier: MIT
+
+# Bricktunes LEGO Color Synth
+# Feather RP2040 Prop-Maker + AS7341 Color Sensor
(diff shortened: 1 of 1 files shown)


CedarGroveStudios:
I was unable to fix the folder problem and submitted an issue for the Community Bundle repo.
Thanks for catching the error!

TheKitty (approved):

CedarGroveStudios:
The Chime library depends on another Community Bundle library, cedargrove_midi_tools. Unfortunately, it appears that circup only recognizes dependencies with `adafruit_...` libraries and core modules listed in `requirements.txt`. I've filed an issue https://github.com/adafruit/CircuitPython_Community_Bundle/issues/192 with the Community Bundle repo and will follow-up with an issue for circup 
In the mean time, the Learning Guide should note that the cedargrove_midi_tools library (available in the bundle and installable with circup) is required.

PR #2651, opened 2023-10-18
fixed  name to avoid confusion

jedgarpark:
I had named the midi object the same as a library, oops! changed object name to `midi_usb` along w `midi_uart` for the other.


TheKitty (approved):
LGTM

PR #2657, opened 2023-11-01
Update Qualia image to work with both types of touchscreens

makermelissa:
This change makes it first try loading as FocalTouch touchscreen, and if that fails, a CST Capacitive touchscreen. Not sure if this makes it too complicated and if I should just go with #if statements, or if you like it as is.

cc: @ladyada

Note: requires Adafruit_CST8XX_Library 1.1.0


ladyada (approved):

ladyada:
this is fine, thanks - did your arduino library get added to the library manager?

makermelissa:
Yes it did, but I made a change this morning so it wouldn't conflict with the FT6206 library and I'm not sure if it's updated yet.

makermelissa:
Ok yeah, it updated to the latest.

PR #2663, opened 2023-11-14
One last Bug fix for messageboard

makermelissa:
This time, it's just fixing a visual glitch instead of a crash.


makermelissa:
Oops, converting to draft. It's currently causing a crash after a long time still.

makermelissa:
Ok, ran it for a while and it looks good.

TheKitty (approved):

PR #2336, opened 2022-11-21
fixed ulab spectrogram imports

evaherrada:
Fixes #2335

PR #2714, opened 2024-01-20
#2713 fixed outdated import of spectrogram

ggsato:
simply fixed the outdated import statement


jepler (approved):
looks fine, didn't test

jepler:
@ggsato please take a look, github says this PR has conflicts. I think we recently incorporated a slightly different fix (#2336) so this may not be needed now. If you can verify that this is the case, please go ahead and close this PR and know that we still appreciate your contribution :)

ggsato:
@jepler Verified. That change is better with the backward compatibility in mind!

PR #2718, opened 2024-01-24
updating example code for the new feather sense

BlitzCityDIY:
updating CP and Arduino examples to check for the LSM6DS33 or the LSM6DS3TR-C

### DIFF
--- a/Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
+++ b/Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
@@ -62,9 +62,7 @@ void setup(void) {
   for (int i =0; i < 5; i++) {
     if (accel_array[i] != check_array[i]) {
       new_rev = false;
-    }
-    else {
-      new_rev = true;
+      break
     }
   }
   // and we need to instantiate the LSM6DS3TRC
--- a/Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
+++ b/Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
@@ -62,7 +62,7 @@ void setup(void) {
   for (int i =0; i < 5; i++) {
     if (accel_array[i] != check_array[i]) {
       new_rev = false;
-      break
+      break;
     }
   }
   // and we need to instantiate the LSM6DS3TRC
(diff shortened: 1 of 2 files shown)


TheKitty (changes requested):
In feather_sense_sensor_demo.ino, two lines where if (new_rev = true) should likely be ==

Also shouldn't booleans be True and False?

BlitzCityDIY:
in arduino booleans are lowercase. i will change to ==.

Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
```diff
@@ -68,7 +68,7 @@ void setup(void) {
     }
   }
   // and we need to instantiate the LSM6DS3TRC
-  if (new_rev = true) {
+  if (new_rev == true) {
```
dhalbert, line 71: Don't need to compare with `true`, just test as `new_rev` and `!new_rev`.
```suggestion
  if (new_rev) {
```

Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
```diff
@@ -95,7 +95,7 @@ void loop(void) {
   sensors_event_t accel;
   sensors_event_t gyro;
   sensors_event_t temp;
-  if (new_rev = true) {
+  if (new_rev == true) {
```
dhalbert, line 98:
```suggestion
  if (new_rev) {
```

dhalbert (changes requested):

Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
```diff
@@ -41,6 +47,30 @@ void setup(void) {
...
+  accel_array[5] = gyro.gyro.z;
+  // if all readings are empty, then new rev
+  for (int i =0; i < 5; i++) {
+    if (accel_array[i] != check_array[i]) {
+      new_rev = false;
+    }
+    else {
+      new_rev = true;
+    }
+  }
```
dhalbert, line 69: This is going to flip-flop `new_rev` back and forth between `true` and `false` as you go through the `for` loop. I'm thinking that's not what you want -- shouldn't it be sticky in one direction or the other (probably in the `false` direction since you start with `true`?)?

dhalbert (changes requested):

Adafruit_Feather_Sense/feather_sense_sensor_demo/feather_sense_sensor_demo.ino
TheKitty: Yes, maybe set it true above the loop and any positive data check can then toggle it false

BlitzCityDIY: good point, adding a break if a value triggers it

TheKitty:
image

PR #2720, opened 2024-01-24
Walkmp3rson

jedgarpark:
there was a memory problem a forum user caught due to mp3 stream method being called repeatedly. tested the fix on hardware, works great.


dhalbert:
@jedgarpark I think you did this PR on a very old fork of the repo. Take a look at all the changes. Merge or rebase from usptream.

PR #2732, opened 2024-02-13
Readability improvements

jacobrose:
## Scope of change

This change is purely for readability.

The LED is disabled in the main loop, so re-enabling it in the main loop makes more sense; also added some spacing between steps to make the code a little more legible.

## Known limitations

None.

## Tests

Unable to test, since I don't have this hardware. Changes should not affect functionality in any way.


TheKitty (changes requested):
image

dhalbert (approved):
Thanks! Formatting issues fixed.

Issue #2746, opened 2024-03-02
DVI_Hello_World - Out of Memory


mikeysklar:
The stock example code for the RP2040 DVI - [CircuitPython dvi-demo or DVI_Hello_World](https://learn.adafruit.com/adafruit-feather-rp2040-dvi?view=all#dvi-demo) is running out of memory. Adjusting the color_depth to 1 which also requires boosting the resolution still results in an out of memory although the code gets another dozen lines further.

[Forum user @Gregg28](https://forums.adafruit.com/viewtopic.php?t=208759) pointed this out. 

I reproduced this errors a Feather RP2040 DVI CP9.0.0-b2:

errors on stock example:
```
Adafruit CircuitPython 9.0.0-beta.2 on 2024-02-20; Adafruit Feather RP2040 DVI with rp2040
>>>
soft reboot

Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
code.py output:
Traceback (most recent call last):
  File "code.py", line 33, in <module>
MemoryError: memory allocation failed, allocating 88332 bytes

Code done running.
```

Adjusting resolution and color depth errors:

```
MemoryError: memory allocation failed, allocating 76800 bytes

Code done running.

Press any key to enter the REPL. Use CTRL-D to reload.
soft reboot

Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
code.py output:
Traceback (most recent call last):
  File "code.py", line 45, in <module>
MemoryError: memory allocation failed, allocating 96000 bytes
```

FoamyGuy:
I've got my Feather DVI out and will attempt to recreate this and dig a bit further into it today.

BlitzCityDIY:
also testing- had 8.2.10 running and haven't run into any problems (let it run for about 45 minutes). running 9 beta2, i ran into the memory allocation issue on first run, hit ctrl+d and it started running:

```
Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
code.py output:
Traceback (most recent call last):
  File "code.py", line 33, in <module>
MemoryError: memory allocation failed, allocating 88332 bytes

Code done running.

Press any key to enter the REPL. Use CTRL-D to reload.
soft reboot

Auto-reload is on. Simply save files over USB to run them or enter REPL to disable.
code.py output:
```

did a hard reset and same behavior, first run failed with memory error, ctrl+d had it run successfully. going to add a try/except and see if that can bypass the issue

[...]

PR #2747, opened 2024-03-04
adding try/except to DVI hello world

BlitzCityDIY:
Adding a try/except to workaround a MemoryError in CircuitPython 9 beta2 as noted in https://github.com/adafruit/Adafruit_Learning_System_Guides/issues/2746

### DIFF
--- a/CircuitPython_DVI_Hello_World/code.py
+++ b/CircuitPython_DVI_Hello_World/code.py
@@ -7,6 +7,7 @@ import gc
 import math
 from random import randint
 import time
+import supervisor
 import displayio
 import picodvi
 import board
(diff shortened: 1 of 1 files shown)


TheKitty:
I do think it'd be good for someone to look at the root cause first before trying to catch the error unless this is a stopgap.

cc @dhalbert @tannewt

FoamyGuy:
I found that the example seems to run successfully on the 1st attempt if the initialization is changed to be like this:
```
# check for DVI Feather with built-in display
if 'DISPLAY' in dir(board):
    display = board.DISPLAY

# check for DVI feather without built-in display
elif 'CKP' in dir(board):
    displayio.release_displays()
    fb = picodvi.Framebuffer(320, 240,
        clk_dp=board.CKP, clk_dn=board.CKN,
        red_dp=board.D0P, red_dn=board.D0N,
        green_dp=board.D1P, green_dn=board.D1N,
        blue_dp=board.D2P, blue_dn=board.D2N,
        color_depth=8)
    display = framebufferio.FramebufferDisplay(fb)
# otherwise assume Pico
else:
    displayio.release_displays()
    fb = picodvi.Framebuffer(320, 240,
        clk_dp=board.GP12, clk_dn=board.GP13,
        red_dp=board.GP10, red_dn=board.GP11,
        green_dp=board.GP8, green_dn=board.GP9,
        blue_dp=board.GP6, blue_dn=board.GP7,
        color_depth=8)
    display = framebufferio.FramebufferDisplay(fb)
```
Basically it will just check for `board.DISPLAY` first and use that if it exists and only if not then attempt to init a display. 

My best guess is perhaps there was some period of time when the device support CircuitPython but didn't have `board.DISPLAY` automatically initialized in the core yet, and that this example (and perhaps more) was created during that time in which it would have been necessary to init the display before showing anything on it.

BlitzCityDIY:
this was early code, right at the release of the board so that makes sense. when i add that check for `board.DISPLAY` i initially got an error for `bitmap = displayio.Bitmap(display.width, display.height, 3)`:

```
  File "code.py", line 53, in <module>
AttributeError: 'NoneType' object has no attribute 'width'
```

i hard coded those (`320, 240`), but then get this error for `display.root_group = group`:
```
  File "code.py", line 331, in <module>
AttributeError: can't set attribute 'root_group'
```

FoamyGuy:
To fix this error:
```
  File "code.py", line 53, in <module>
AttributeError: 'NoneType' object has no attribute 'width'
```
You need to remove the `displayio.release_displays()` from the top of the script. It should only be inside of the elif and else now. (Sorry I didn't mention that in my prior post).

I think the second error is likely related to this as well, but I'm not as certain on that one. 

Make this change to remove that line from the top, then unplug / re-plug once after you've saved the new code and see if you still have any errors after that.

FoamyGuy:
I tested this PR version succesfully on the Feather DVI with 9.0.0 beta.2

We can still let Scott or Dan chime in for final approval, but this looks good to me.

The fact that after `release_displays()` there is not enough memory to re-initialize the display sounds like a potential bug to me, but I'm not certain of it. And if it is, I do think it's a separate issue and this fix does still stand on it's own as an improvement.

tannewt:
> I do think it'd be good for someone to look at the root cause first before trying to catch the error unless this is a stopgap.
> 
> cc @dhalbert @tannewt

Thanks @TheKitty. It was a bug in CP. It is good to just not work around it.

> We can still let Scott or Dan chime in for final approval, but this looks good to me.

I think this change to use `board.DISPLAY` when available is good. Any reloading would be suspect to me.

> The fact that after `release_displays()` there is not enough memory to re-initialize the display sounds like a potential bug to me, but I'm not certain of it. And if it is, I do think it's a separate issue and this fix does still stand on it's own as an improvement.

Yup, it was. The next 9.0.0 release should have a fix for it. https://github.com/adafruit/circuitpython/pull/9004

BlitzCityDIY:
just to check- i know this ended up being a CP issue but the change here to check for `board.DISPLAY` now that that is in the Feather DVI board def should still be added for this example?

FoamyGuy (approved):
Yes, it should get this check for sure. Scott mentioned that as well. 

There is no reloading left in the code either. This change looks good to me. Thanks for the fix @BlitzCityDIY

PR #2752, opened 2024-03-16
make feather can ack example work with stm32f405

Kheirlb:
  the code.

Avoids the issue in https://github.com/adafruit/circuitpython/issues/3644

The STM32F405 Feather differes from the other adafruit CAN friendly boards because it has 3 "transmit mailboxes". The other boards seem to have only 1 tx buffer (not 100% sure about this). The current ack example does not work well on the stm boards after bus errors because it assumes the next received message will be after the most recent sent message, which is not the case because of these 3 transmit mailboxes. For example:

1. send 1, receive 1
2. send 2, receive 2
3. *can bus disconnection*
4. send 3, buffer (tx mailbox0)
5. resend 3, buffer (tx mailbox1)
6. resend 3, buffer (tx mailbox2)
7. resend 3, buffer (out of mailboxes, replaces value in mailbox0)
8. *can bus reconnection*
9. recieves 3! behind the scenes, the stm transmits the remaining 2 messages in the transmit buffer
10. send 4, receives 3 (oops! this was a 3 in the buffer)
11. resend 4, receives 3 (oops! this was the last 3 in the buffer)
12. resend 4, receives 4 (yay! but this is the wrong one, and it was actually the first 4 from step 10)
13. send 5, receives 4 (oops! that is the second 4... process continues...)

This is very clear in the screenshot here https://github.com/adafruit/circuitpython/issues/3644#issuecomment-1982235235

I am worried the iterator over the listener may be a little confusing to the novice ciruitpython user. This is the easier way though for the CiruitPython loop to _consume_ all of those "pending" transmit attempts before moving on.

  process.

I ran this locally and it seems to work well.

### DIFF
--- a/FeatherCAN_CircuitPython/listener-ack/code.py
+++ b/FeatherCAN_CircuitPython/listener-ack/code.py
@@ -50,5 +50,5 @@ while True:
     if gap != 1:
         print(f"gap: {gap}")
 
-    print("Sending ACK")
+    print(f"Sending ACK: {count}")
     can.send(canio.Message(id=0x409, data=struct.pack("<I", count)))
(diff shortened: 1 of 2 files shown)


dhalbert:
@jepler mind reviewing this?

Kheirlb:
Friendly ping. Can someone with write access merge this?

TheKitty (approved):

PR #2753, opened 2024-03-20
Picowbell doubler battery monitor code

BlitzCityDIY:
adding PiCowbell Doubler battery monitor code for CircuitPython and Arduino


TheKitty:
Rerunning jobs that failed due to dependency upload failure (not to do with the code for this project)

PR #2647, opened 2023-10-16
displayio api updates for 9.x+

FoamyGuy:
This supersedes https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/2429

I've begun the process of updating code to use the new displayio API. 

As I'm working through these I'm also checking in the Learn Guides and making a seperate list of ones that have non-github embedded code in the guide that will need to be changed.

PR #2756, opened 2024-03-23
replace .show(x) with .root_group = x

dhalbert:
Update all the code I could find that needed .`show(...)` to be replaced with `.root_group = ...`

I also did a Google site search to find as many places as I could that needed this fix in the Guides text.

### DIFF
--- a/CLUE_Light_Painter/code.py
+++ b/CLUE_Light_Painter/code.py
@@ -230,7 +230,7 @@ class ClueLightPainter:
             board.DISPLAY.root_group = group
             sleep(4)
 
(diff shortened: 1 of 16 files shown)


FoamyGuy (approved):
These look good to me. 

Many of these were taken care of by https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/2647 but some must have been missed or added since.

dhalbert:
> Many of these were taken care of by #2647 but some must have been missed or added since.

I thought so too, but I think I got the remaining ones now. There are some `show()` which are unrelated (especially calls with no argument, which I ignored).

PR #2758, opened 2024-03-25
User must provide a mount_sd.py script

jepler:
this import line was removed without checking, because it caused a pylint message. (24dacb7fe6c5d4702b8b81c59cfcea0a7b7ce416) -- the guide may need to be reviewed as well, to make sure usage of mount.py is explained (I'll tackle this next)


jepler:
the guide looks OK, the usage of mount_sd.py is mentioned in multiple spots. This just restores the lines that were removed and adds a comment to explain it

TheKitty (approved):

PR #2767, opened 2024-04-09
PyPortal_LIFX_Controller: get LIFX_TOKEN from settings.toml

dhalbert:
Updating project to use settings.toml


brentru (approved):
@dhalbert Thanks for migrating this code.py. Changes look correct for using settings.toml instead of secrets.py

TheKitty (approved):

PR #2768, opened 2024-04-09
Update PyPortal_Philips_Hue_Controller to use settings.toml

dhalbert:


brentru (approved):
Addresses the move to secrets.toml

PR #2775, opened 2024-04-09
Update TOTP_Friend secrets.py

dhalbert:
This is part of updating the TOTP_Friend project to settings.toml. The wifi settings are now in settings.toml, but the keys reamain in secrets.py, because they are in a hierarchical structure.

By mistake, I committed the code.py changes directly instead of making a PR.

Guide already updated.


brentru (approved):
Removal of the wifi secrets and tz look correct. Will leave it to you to merge when pylint is fixed @dhalbert

PR #2777, opened 2024-04-09
PyPortal_Quarantine_Clock: update to settings.toml

dhalbert:
Guide also updated.

### DIFF
--- a/PyPortal/PyPortal_Quarantine_Clock/code.py
+++ b/PyPortal/PyPortal_Quarantine_Clock/code.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: MIT
 
+import os
 import time
 import board
 import busio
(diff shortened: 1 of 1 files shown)


PyPortal/PyPortal_Quarantine_Clock/code.py
```diff
@@ -88,7 +82,7 @@
     if (not refresh_time) or (time.monotonic() - refresh_time) > 3600:
         try:
             print("Getting new time from internet...")
-            pyportal.get_local_time(secrets['timezone'])
+            pyportal.get_local_time(os.getenv("TIMEZONE"))
```
brentru, line 85: The timezone was previously lower case "timezone". 

In the guide - Did it change to "TIMEZONE" in the toml file?

brentru (changes requested):

PyPortal/PyPortal_Quarantine_Clock/code.py
dhalbert: Yup, guide is all upper-case entries in settings.toml now.

dhalbert:
The build failure is in an unrelated file. Ready to re-review.

TheKitty (approved):
image
You defined OS but it's not finding. Approved

dhalbert:
Yeah, that error  is in TOTP_Friend, which I forgot to say "Make a new branch on". I fixed that, but it is running the PR tests against the pre-fixed version. Problem will go away on merge.

PR #2790, opened 2024-04-20
Move app name a few pixel down for better vertical alignment

ViToni:
  of the code were modified.**
For the MacroPad Hotkeys project the "app name" is touching the top of the header bar.
As there is "enough" space this change makes the header bar one pixel higher and moves the text 2 pixels down.
This tiny changes improves the vertical aligment of the "app name". 

- **Describe any known limitations with your change.**
Applies only to the "MacroPad Hotkeys project / code".


dhalbert (approved):
Thanks!

PR #2815, opened 2024-05-20
picowbell tripler demos

BlitzCityDIY:
CircuitPython and Arduino demos for the PiCowbell Tripler


TheKitty (approved):
I'd suggest future code have a few more comments to help readers know what's what

PR #2841, opened 2024-07-01
LED Bullwhip update for Feather Sense LSM6DS

mikeysklar:
The Feather Sense used in this project has two different versions. The LSM6DS changed in January 2024 from the LSM6DS33 to the LSM6DS3TR-C. They have slightly different initialization. The guide code.py has been modified to support both versions. 

This has been tested on two of the new versions of the Feather Sense (new models) by two people and shown to run.


ladyada:
thanks!

PR #2926, opened 2024-11-18
arduino simpletest for the vcnl4200

BlitzCityDIY:
arduino simpletest for the vcnl4200. reads lux, white light and proximity data


Arduino_VCNL4200_simpletest/Arduino_VCNL4200_simpletest.ino
```diff
@@ -0,0 +1,48 @@
...
+
+void loop() {
+  uint16_t proxData = vcnl4200.readProxData();
+  Serial.print("Prox Data: ");
+  Serial.println(proxData);
+  // Read the ambient light sensor (ALS) data
+  uint16_t alsData = vcnl4200.readALSdata();
+  Serial.print("ALS Data: ");
+  Serial.print(alsData);
+  uint16_t whiteData = vcnl4200.readWhiteData();
```
brentru, line 43: Add comment:
`// Read the raw white sensor data` 

to match comment on L39

brentru (approved):
@BlitzCityDIY The refactor for FastLED looks good. I left one optional comment, you may accept or reject it as it does not affect the program itself.

LGTM

BlitzCityDIY:
@brentru awesome, thanks! that makes sense about the comment, will add.

brentru:
Sounds good - comments like this make the loop code appear more symmetrical.

GemmaM0_Band_Jacket/DiscoBandCamp/effects.h
```diff
@@ -286,7 +286,7 @@ void swirly()
   // blur it repeatedly.  Since the blurring is 'lossy', there's
   // an automatic trend toward black -- by design.
   uint8_t blurAmount = beatsin8(2,10,255);
-  blur2d( leds, kMatrixWidth, kMatrixHeight, blurAmount);
```
jepler, line 289: I know we're both just making guesses here but I think this is going to change the result of blur2d.

I went back into the fastled history to see how this worked before and after:
https://github.com/FastLED/FastLED/commit/b02a481a294740196a0532a0ab93d706e6d413fc#diff-6b91d9b6efd432b06dc6eb73a57f6a29d64d6b6983c760779f9e86099d8ddafdL465

Previously, blur2d used the XY function provided in the sketch, this was implicit. So the blur would be using the structure of the disco jacket as defined in the Jacket Table.

After this change, the XY function is not used by blur (though it is used elsewhere through the file), but instead a serpentine mapping defined by myXYMap would be used.

I think that [`constructWithUserFunction()`](https://fastled.io/docs/class_x_y_map.html#a2887ddf6bd89bbcbe4ab618e6923170b) probably needs to be used, something like: `XYMap myXYMap = XYMap::constructWithUserFunction(kMatrixWidth, kMatrixHeight, XY)`

The result of an incorrect pixel map is that values from one area of the jacket would blur into non-adjacent pixels, making the effect look more muddled.

jepler (commented):
I looked into the XY mapping stuff a bit. I didn't find a migration guide but I have some guesses that the change made here is not identical in functionality to the old code.

GemmaM0_Band_Jacket/DiscoBandCamp/effects.h
BlitzCityDIY: very good catch and i was concerned about that one. i just pushed an update that uses [constructWithUserFunction()](https://fastled.io/docs/class_x_y_map.html#a2887ddf6bd89bbcbe4ab618e6923170b). i hooked up two neopixel matrices and a button to test and compare to the project video and i believe that functionality has been retained

jepler:
Thanks!

PR #2948, opened 2025-01-13
tmc2209 examples

FoamyGuy:
cc @BlitzCityDIY 
Almost the same as A4988 ones, but different default microstep value.


BlitzCityDIY:
thank you!

PR #2951, opened 2025-01-22
adding arduino sparkle motion mini example

BlitzCityDIY:


BlitzCityDIY:
something broke with rp2040 and sdcard, will investigate

PR #2960, opened 2025-02-02
quick fix broken imports/calls of pcf8523

semininja:
Several example files incorrectly `import adafruit_pcf8523` and do `rtc = adafruit_pcf8523.PCF8523(I2C)` when they should `from adafruit_pcf8523.pcf8523 import PCF8523` and `rtc = PCF8523(I2C)`. This PR corrects these imports and makes no other changes. I don't own the various bits of hardware necessary to test these examples, but my changes here *should* have no other impact.

### DIFF
--- a/Daily_Cheer_Automaton/CircuitPython/code.py
+++ b/Daily_Cheer_Automaton/CircuitPython/code.py
@@ -40,7 +40,7 @@ import busio as io
 # import bitbangio as io
 
 # import adafruit_ds3231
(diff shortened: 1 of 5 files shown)


caternuson:
Thanks. These LGTM in terms of syntax. As an additional reference, these updates are consistent with what was done in this similar PR:
https://github.com/adafruit/Adafruit_Learning_System_Guides/pull/2722

PR #2962, opened 2025-02-04
fixed issue of reading Network object in code.py of example for Arduino Nano …

edenchiang:
…RP2040 Connect

### Description
This is a fix for few syntex errors and also an update for the example according to the latest documentation.

### Working environment: 
CircuitPython version: 9.2.4

### Target platform
Arduino Nano RP2040 Connect

#### Line 49:
The `ap` is a `Network` object which is not in dictionay type, so the content should be retrieved by `obj.attribute_name`

#### Line 58:
The `esp` object has no dictionary keys named `ssid `and `rssi`, but which are packed in the `ap_info` object according to the official documentation: [adafruit_esp32spi - ap_info](https://docs.circuitpython.org/projects/esp32spi/en/latest/api.html#adafruit_esp32spi.adafruit_esp32spi.ESP_SPIcontrol.ap_info)


dhalbert (approved):
Thanks for fixing these! This obsolete code was fixed elsewhere, but I didn't know there was a separate example for the Nano RP2040 Connect.

PR #2967, opened 2025-02-11
adding cartoon character clock

FoamyGuy:


TheKitty:
image

FoamyGuy:
I've re-arranged the imports in accordance with that rule. 

The other pylint errors I've disabled because the if statement logic that already wraps the usage of relavent variables prevents use before assignment, and accessing unsubscriptable values. 

This code cannot use `with:` for the file openings because the file must remain open for longer than 1 context block.

FoamyGuy:
@TheKitty I think this one is ready for review now.

Some of those pylint options were only in my local version which is a bit newer than the one used by actions here. The latest commit leaves only the ones present in the versions that the check uses.

TheKitty (approved):

PR #2973, opened 2025-02-18
Arduino SD Card example

FoamyGuy:


TheKitty:
@FoamyGuy did you make a .test.only file?

FoamyGuy:
@TheKitty I've added `.none.test.only` with the latest commit. As far as I can tell there is no Metro 2350, or any other 2350 based devices listed in the jobs. This failed every platform it tried on the first time.

If there is a metro2350 or other 2350 one added in the future this file could be updated to point to it instead of `.none` if my understanding of how the tests work is correct.

TheKitty:
Yes, the process is at https://learn.adafruit.com/submitting-arduino-code-to-the-adafruit-learning-system/adding-a-test-only-file

Yes, the file https://github.com/adafruit/ci-arduino/blob/master/all_platforms.py does not have the definition for a metro_rp2040 or a metro_rp2350

TheKitty (approved):
Please set a task to revisit this once the board definition is available to change the .test.only filename

FoamyGuy:
I've added a todo in basecamp for it.

PR #2976, opened 2025-02-19
USB Host gamepad read data example

FoamyGuy:
Example for reading data from a generic USB game pad on the Feather RP2040 USB Host


TheKitty (approved):
Nice code!

PR #2998, opened 2025-03-24
into the jammy matrix

jepler:
A version of @TheKitty 's HSTX matrix scroller that shows the message "Adafruit Fruit Jam"

Added as a new sketch per request by @ladyada


ladyada:
you can add a .metro_rp2350.test.only file

jepler:
@ladyada I think the remaining failures are not due to my change, but they also don't seem to affect other recent PRs so I'm not sure what's up.

jepler:
thanks

PR #3007, opened 2025-04-01
Add Chips Challenge for guide

makermelissa:
Here's the code in its entirety needed for my guide.


makermelissa:
Thanks. I decided to not do it and instead provide it as an option to users.

TheKitty (approved):
LGTM

PR #3018, opened 2025-04-11
Detect max quality level dynamically.

casten:
Detect max quality level dynamically.
This will select 2 for standard Pico 2's and 5 for standard Pico's.
Tested on Pico and Pico 2.

Note that the the code for Adafruit_CircuitPython_OV5640, the acceptable range of the quality says:
`        """Controls the JPEG quality.  Valid range is from 2..55 inclusive"""`
However, the value 55 is not included, so grammatically I'd say this is inaccurate. 
https://github.com/jerryneedell/Adafruit_CircuitPython_OV5640/blob/main/adafruit_ov5640/__init__.py#L1495
 I've adjust my comment in the code to indicate such.


BlitzCityDIY (approved):
great, i like this approach

PR #3020, opened 2025-04-14
Update - WS Offline Guide

brentru:
This pull request updates the `config.json` file for the following guide: https://learn.adafruit.com/no-code-offline-data-logging-with-wippersnapper/usage

The new `config.json` file reflects a file that is automatically generated by the WipperSnapper Offline Beta 2 release.

### DIFF
--- a/WipperSnapper_Offline/config.json
+++ b/WipperSnapper_Offline/config.json
@@ -1,25 +1,36 @@
 {
     "exportedFromDevice": {
-        "referenceVoltage": 3.3,
-        "totalGPIOPins": 18,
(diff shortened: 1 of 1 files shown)


TheKitty (changes requested):
Are you sure about these? 
      "referenceVoltage": 0,
      "totalGPIOPins": 0,

brentru:
@TheKitty  Yep - that's a good catch. But those values are zero'd out by default but will be initialized later when the user uses the configuration tool.

TheKitty (approved):

PR #3023, opened 2025-04-17
USB SNES-like controller example for arduino

FoamyGuy:


TheKitty (changes requested):
Spelling: If already enumerated, additional class driverr

PIO USB require CPU clock must be multiple of 120 Mhz - is that true or 12 MHz?

FoamyGuy:
> PIO USB require CPU clock must be multiple of 120 Mhz - is that true or 12 MHz? 

I think it's 120 Mhz, but honestly I am not 100%, nor do I know the reason for the limitation. The code here seems to enforce it though: https://github.com/FoamyGuy/Adafruit_Learning_System_Guides/blob/7edbbf2401b81734db129807f714c056cba07048/USB_SNES_Gamepad/Arduino_USB_Host/usbh_helper.h#L68

I reused that file from here: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/USB_Host_Turbo_Button_Gamepad/Arduino_USB_Host_Turbo_Button_Gamepad/usbh_helper.h

TheKitty:
> > PIO USB require CPU clock must be multiple of 120 Mhz - is that true or 12 MHz?
> 
> I think it's 120 Mhz, but honestly I am not 100%, nor do I know the reason for the limitation. The code here seems to enforce it though: https://github.com/FoamyGuy/Adafruit_Learning_System_Guides/blob/7edbbf2401b81734db129807f714c056cba07048/USB_SNES_Gamepad/Arduino_USB_Host/usbh_helper.h#L68
> 
> I reused that file from here: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/USB_Host_Turbo_Button_Gamepad/Arduino_USB_Host_Turbo_Button_Gamepad/usbh_helper.h

Please check with Scott, I was thinking this was an evolving thing

TheKitty (approved):

PR #3024, opened 2025-04-17
add support for Metro rp2350 USB SNES-like controller

FoamyGuy:
Also:
- removed USB HID device handling code that was unused by this project. 
- changed to use a single core & setup/loop functions
- added prints during setup to output which pins are in use for USB


TheKitty:
Some comments still refer to two cores 
For rp2040 pico-pio-usb, calling USBHost.begin() on core1 will have most of the
  // host bit-banging processing works done in core1 to free up core0 for other works

FoamyGuy:
Thank you, the reference to different cores are removed with the latest commit.

TheKitty (approved):

PR #3025, opened 2025-04-22
Usb host mouse examples

FoamyGuy:
I created a new top level director `USB_Host_Examples/` and put these mouse demos inside of it. I am working on the USB Host keyboard guide next and plan to put the keyboard examples in this new folder as well.  In the future theoretically some of the older USB host projects could be moved into this folder if we wanted as well, in order to consolidate things outside of the root directory.

PR #3026, opened 2025-04-23
usb host keyboard examples

FoamyGuy:
This is in the same new top level directory `USB_Host_Examples/` as the examples from #3025.

Each example lives within its own sub directory with a specific name, so I don't think there should be merge conflicts when one of these PRs is merged for the other one, but if any do pop up I'll add a commit to resolve them.


FoamyGuy:
The latest commit disables the pylint rule inside of Chips Challenge code which was causing the actions here to fail.

TheKitty (approved):

PR #3034, opened 2025-05-10
Tilepalettemapper api update

FoamyGuy:
update code for the new TilePaletteMapper API from https://github.com/adafruit/circuitpython/pull/10318

Also updated the CircuitPython Matrix code to request the smaller display resolution so that it draws fewer total characters and refreshes faster.


TheKitty (commented):
Is this project to only be run under CP 10? Should it be 9 and 10 compatible?

TheKitty (approved):

FoamyGuy:
I added info alerts on the relavent pages in the Match3 guide noting the active work on USB and suggesting to upgrade if they have trouble with mice 
image

PR #3044, opened 2025-05-29
fix for sparkle motion mini warning

FoamyGuy:
I believe this change should resolve this actions failure resulting from a warning about comparing signed with unsigned types: https://github.com/adafruit/Adafruit_Learning_System_Guides/actions/runs/15316694255/job/43091814706#step:7:902

The issue boils down to `Adafruit_Neopixel.numPixels()` returning an unsigned value, but `int i = 0` from the loop variable being signed. 

I tested the modified version successfully on the sparkle motion mini with two 1x8 NeoPixel strips.


FoamyGuy:
The new commit has the same fix for the big sparkle motion, I also tested the updated code on that hardware successfully. 

I'm unclear about why the GH actions seem to flag this as a failure under gemma and put the X next to `Starduino_8bit_Super_Mario_Tree_Topper.ino`. It seems like maybe the thing coming after the warning failure ends up being flagged as the failing task instead of the one that caused the warning.

ladyada (approved):

PR #3047, opened 2025-06-01
fix functionality for RP2x hosts

ladyada:
tested on fruit jam (ESP32-C6 client)


brentru (approved):
@ladyada Looks good, I can see why you'd want to switch to a LED from NeoPixel for this board. May be useful in the future to expand the arch define for rp2040 to other boards that don't contain a status NeoPixel/DotStar.

brentru:
I don't have a Fruitjam so I am unable to test on hardware. You already tested on this h/w, so feel free to merge this PR at your convenience.

PR #3051, opened 2025-06-03
feather s2 bme280 deepsleep aio example

FoamyGuy:
Arduino and CircuitPython examples for Feather S2 with BME280 sensor to deep sleep, wake up and post data to AIO.


FoamyGuy:
I am not sure what is up with the actions failures. It seems like they are for things unrelated to the new examples on devices that aren't the feather s2. 

Maybe `.feather_esp32s2.test.only` is not the right name for the file to go alongside the new examples? I copied that filename from this other one I found in the repo: https://github.com/adafruit/Adafruit_Learning_System_Guides/tree/main/Factory_Tests/Feather_ESP32S2_FactoryTest

FoamyGuy:
This is passing actions now due to some fixes submitted elsewhere.

I've taken another pass over the code and expanded and clarified some of the comments a bit and I think this is ready to go now.

PR #3062, opened 2025-06-13
update sd examples for 10.x behavior

FoamyGuy:
Adds try/except for initializing the SD to account for new 10.x feature that the core will auto initialize it. 

I tested the updated examples with both 9.2.8 and 10.0.0-alpha.6 on Metro RP2350


TheKitty (commented):
I think the code could use some more comments for clarity

TheKitty (approved):

PR #3063, opened 2025-06-13
code for not a typewriter

BlitzCityDIY:
both versions of the not a typewriter code. one uses only circuitpython with usb host for a usb keyboard. the second has a desktop python script that communicates with the feather over serial


TheKitty (approved):
Great comments

PR #3064, opened 2025-06-17
adding some icons and metadata for fruitjam os

FoamyGuy:
These icon and metadata.json files get used by the fruit jam OS launcher.

Other changes include:
 -  swapping a few projects to use the adafruit_fruitjam library to request a display size instead of assuming the appropriate one, or manually initializing it.
 - change the behavior of exiting in a few cases to integrate with the fruit jam OS launcher better.
 - use adafruit_usb_descriptors helper to find appropriate mouse USB interface instead of assuming the first one found is correct.

### DIFF
--- a/Metro/Metro_RP2350_FlappyNyanCat/code.py
+++ b/Metro/Metro_RP2350_FlappyNyanCat/code.py
@@ -13,7 +13,7 @@ Controls:
 import random
 import sys
 import terminalio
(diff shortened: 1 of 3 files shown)


TheKitty (commented):
@FoamyGuy - would it be better that the request_display_config routine be in displayio rather than a Fruit Jam library when it might be used in non-Fruit Jam projects?

PR #3085, opened 2025-07-17
HXTX DVI CowBell Arduino Examples

FoamyGuy:
- moved circuitpython examples to a subdirectory
- added arduino examples for both rp2040 with PicoDVI, and rp2350 with HSTX_DVI

I'm not sure if `.pico_rp2040.test.only` is the correct file for the rp2350 example. I didn't see any file like `.pico_rp2350.test.only` in use anywhere else in this repo so I assumed that might not work.


FoamyGuy:
Some or all of this may not be needed. HSTX examples are being updated in the library repo to choose the correct pin config for the device they are on which makes the example more universal and it can be embedded in the guide pages from the library repo I think. 

I will either close, or make changes and take this back out of draft once I've got a grasp on whether we need any parts of this.

TheKitty (approved):

FoamyGuy:
Removed 2350 example, it will get embedded from HSTX-DVI repo which has logic to select correct pin config. 

I looked into doing the same for 2040 but I think it's not as feasible for that one because the choice between hstx dvi cowbell and older dvi cowbell isn't possible to know at runtime from the code. The user must select the one that matches their hardware.

PR #3089, opened 2025-08-01
first commit esp-now basics

jedgarpark:
esp-now basics guide transciever code

### DIFF
--- a/ESP-NOW_basics/espnow_transceiver.py
+++ b/ESP-NOW_basics/espnow_transceiver.py
@@ -1,0 +1,157 @@
+# SPDX-FileCopyrightText: John Park for Adafruit 2025
+# SPDX-License-Identifier: MIT
+"""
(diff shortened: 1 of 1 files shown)


TheKitty (approved):

dglaude:
Dear @jedgarpark,

Your status update after 0.2s make that the error message is very short lived.
Also, every loop turn you update the status_label.txt and .color to the same value.
This constant update is not visible because you are in graphic mode, but not very efficient.
To convince myself, I added something like that `print("I was here")`:

```
        except Exception as ex: # pylint: disable=broad-except
            print(f"Send failed: {ex}")
            status_label.text = "xxx"
            status_label.color = RED

        last_button_time = current_time

    # Reset status after a moment
    if current_time - last_button_time > 0.2:
        status_label.text = "press D0 to send"
        status_label.color = TOMATO
#        print("I was here")

...

    time.sleep(0.05)  # Light polling
```

PR #3103, opened 2025-08-09
Add audio output option file to Larsio and Chip's Challenge

RetiredWizard:
This Updates both the Larsio_Paint_Music and Chip's Challenge learning guides.

This PR adds the logic necessary to read a json configuration file (launcher.conf.json, if it exists in either the current or root folders) which has an option to direct the DAC audio output to the Speaker connection rather than the default headphones. This replicates the parameter added to the Fruit Jam OS boot_animation via https://github.com/adafruit/Fruit-Jam-OS/pull/49.

The Json syntax for the parameter is as follows:
```
{
    "sound": "speaker"
}
```
If the launcher.conf.json file is not found, the parameter is not found in the file or it is not set to "speaker" the programs will fall back to the default headphones output option. The Chip's Challenge program has an additional option of "mute" which will disable the audio output potentially avoiding a crash is no audio hardware is available.

### DIFF
--- a/Fruit_Jam/Larsio_Paint_Music/sound_manager.py
+++ b/Fruit_Jam/Larsio_Paint_Music/sound_manager.py
@@ -9,6 +9,7 @@
 import math
 import time
 import array
+import json
 import gc
 import os
 import digitalio
@@ -22,6 +23,7 @@ import audiobusio
 import audiomixer
 import synthio
 import board
+import adafruit_pathlib as pathlib
 import adafruit_tlv320
 from adafruit_midi.note_on import NoteOn
 from adafruit_midi.note_off import NoteOff
@@ -63,6 +65,15 @@ class SoundManager:
             # Setup PWM audio output on D10
             self.audio = audiopwmio.PWMAudioOut(board.D10)
         else:  # i2s
+            # optional configuration file for speaker/headphone setting, check current and root directory
+            launcher_config = {}
+            if pathlib.Path("launcher.conf.json").exists():
+                with open("launcher.conf.json", "r") as f:
+                    launcher_config = json.load(f)
+            elif pathlib.Path("/launcher.conf.json").exists():
+                with open("/launcher.conf.json", "r") as f:
+                    launcher_config = json.load(f)
+
             try:
                 # Import libraries needed for I2S
                 #check for Metro RP2350 vs. Fruit Jam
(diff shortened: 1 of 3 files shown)


Fruit_Jam/Larsio_Paint_Music/sound_manager.py
```diff
@@ -96,20 +108,51 @@ def __init__(self, audio_output="pwm", seconds_per_eighth=0.25):
...
-
-                # Setup I2S audio output - important to do this AFTER configuring the DAC
-                self.audio = audiobusio.I2SOut(
-                    bit_clock=bclck_pin,
-                    word_select=wsel_pin,
-                    data=din_pin
-                )
-
-                print("TLV320 I2S DAC initialized successfully")
+                if ltv320_present:
```
FoamyGuy, line 123: I think it's fine to leave initialization of the TLV320 unguarded by the if statement and above logic checking for it. 

This project specifies Fruit Jam in the name and the TLV is built-in so safe to assume it's there. The original Metro version that shown in the guide also uses the TLV320 so I think it's fine for the code to go ahead and init without checking.

Since this project is sound based and documented with the TLV I think it's best to have it error out if the TLV is not found rather than try to fall back to a different audio type that isn't documented anywhere.

Metro/Metro_RP2350_Chips_Challenge/code.py
```diff
@@ -39,16 +38,12 @@
 
 displayio.release_displays()
 
-i2c = board.I2C()
-dac = adafruit_tlv320.TLV320DAC3100(i2c)
```
FoamyGuy, line 43: I lean towards leaving the hardware init and configuration in code.py rather than move it to definitions.py. I would defer to @makermelissa if she has an opinion on it.

Fruit_Jam/Larsio_Paint_Music/sound_manager.py
RetiredWizard: makes sense, I'll hop on this later tonight.

Metro/Metro_RP2350_Chips_Challenge/code.py
RetiredWizard: It would be easy enough to move the init back to code.py. The reason I moved it to definitions.py was that I thought it was useful for setting a default value for PLAY_SOUNDS that would avoid crashes on hardware that didn't have the DAC hardware. The sound in Chip's Challenge is awesome, but the game is still lots of fun without it.

The default isn't necessary though and after looking at definitions.py again I can see that my code made it less clear that PLAY_SOUND was a definition that could be customized.  I've added an update that I think helps with that issue but I'm happy to switch the default back to True and move the initialization back to code.py if that's the way you and Melissa want to go :grin:

Fruit_Jam/Larsio_Paint_Music/sound_manager.py
```diff
@@ -98,9 +106,24 @@ def __init__(self, audio_output="pwm", seconds_per_eighth=0.25):
 
                 # Initialize TLV320
                 self.tlv = adafruit_tlv320.TLV320DAC3100(i2c)
+                self.tlv.reset()
```
FoamyGuy, line 109: Does the TLV320 need to get `reset()` here?  It should be trying to reset during the init function here: https://github.com/adafruit/Adafruit_CircuitPython_TLV320/blob/f24788d94ad5288fd26ba9b01ded8107ba943c86/adafruit_tlv320.py#L866-L867 Were you seeing issues with it that led to adding the reset() here?

Fruit_Jam/Larsio_Paint_Music/sound_manager.py
```diff
@@ -98,9 +106,24 @@ def __init__(self, audio_output="pwm", seconds_per_eighth=0.25):
...
+                        self.tlv.speaker_output = True
+                        self.tlv.dac_volume = launcher_config["tlv320"].get("volume",5)  # dB
+                    else:
+                        # use headphones
+                        self.tlv.headphone_output = True
+                        self.tlv.dac_volume = launcher_config["tlv320"].get("volume",0)  # dB
+                else:
+                    # default to headphones
+                    self.tlv.headphone_output = True
+                    self.tlv.dac_volume = 0  # dB
```
FoamyGuy, line 126: does this end up playing louder than the previous sound level with `-15` on headphone_volume?

FoamyGuy (commented):
A few more questions.

Fruit_Jam/Larsio_Paint_Music/sound_manager.py
RetiredWizard: I added that when I was having problems controlling the volume before learning about dac_volume. A reset in the init should be sufficient, I'll get rid of this one.

Metro/Metro_RP2350_Chips_Challenge/code.py
RetiredWizard: Another reason for moving the sound config to definitions.py was that although it's probably unlikely, if @makermelissa ever wanted to make some of the parameters in definitions.py user settable, the launcher_config dictionary would already be available without having to have it opened and read in two different modules.

FoamyGuy (approved):
This looks good to me. 

I tested the updated apps successfully on both Fruit Jam and Metro RP2350 + TLV320 on CircuitPython 10.0.0-beta.3 volume controls from the config file are working nicely for both.

Issue #3116, opened 2025-09-02
Arduino CI failures with FastLED library


BlitzCityDIY:
https://github.com/adafruit/Adafruit_Learning_System_Guides/actions/runs/17393839991

PR #3117, opened 2025-09-02
arduino CI fix for fastled

BlitzCityDIY:


TheKitty (approved):

BlitzCityDIY:
fixes https://github.com/adafruit/Adafruit_Learning_System_Guides/issues/3116 for now. the fastled 3.10.2 release is major and refactors how sketches are compiled with the library. for now i have pinned fastled to 3.10.1 and added an error raise to all fastled sketches if it detects a version above 3.10.1

PR #3121, opened 2025-09-18
new float volume setting and don't set volume to override

RetiredWizard:
@FoamyGuy 

This updates the default volume settings to the float 0.0-1.0 range and restricts the "volume_override_danger" parameter to setting only the safe_volume_limit and not directly impacting the volume setting.

I'm marking this as draft as the merge should be coordinated with other apps and library PRs including https://github.com/adafruit/Adafruit_CircuitPython_FruitJam/pull/22


FoamyGuy (approved):
I think this is good to go.

I pushed a commit to update the usage inside of spell jam app as well. I believe that is the only other current usage of the API in learn projects.

This should get merged at the same time as these once they've been reviewed:
https://github.com/adafruit/Adafruit_CircuitPython_FruitJam/pull/22
https://github.com/adafruit/Fruit-Jam-OS/pull/71

PR #3122, opened 2025-09-18
Larsio_Paint_Music: refactor for use with adafruit_usb_host_mouse lib…

RetiredWizard:
This PR removes some of the low level mouse initialization and processing from the Larsio_Paint_Music application and replaces it with the API from the Adafruit_USB_Host_Mouse library.

https://github.com/adafruit/Adafruit_CircuitPython_USB_Host_Mouse/pull/5 should probably be merged before this one.

### DIFF
--- a/Fruit_Jam/Larsio_Paint_Music/input_handler.py
+++ b/Fruit_Jam/Larsio_Paint_Music/input_handler.py
@@ -7,6 +7,8 @@
 
 import array
 import gc
+
+# pylint: disable=import-error
 from adafruit_usb_host_mouse import find_and_init_boot_mouse
 
 # pylint: disable=invalid-name,no-member,too-many-instance-attributes,too-many-arguments
--- a/Fruit_Jam/Larsio_Paint_Music/input_handler.py
+++ b/Fruit_Jam/Larsio_Paint_Music/input_handler.py
@@ -28,11 +28,9 @@ class InputHandler:
         self.buf = None
         self.in_endpoint = None
 
-        self.sensitivity = 4  # Sensitivity factor for mouse movement
-
         # Mouse position
-        self.mouse_x = (screen_width // 2) * self.sensitivity
-        self.mouse_y = (screen_height // 2) * self.sensitivity
+        self.mouse_x = screen_width // 2
+        self.mouse_y = screen_height // 2
 
     def find_mouse(self):
         self.mouse = find_and_init_boot_mouse(cursor_image="sprites/lars_note.bmp")
@@ -41,8 +39,7 @@ class InputHandler:
             return False
 
         # Change the mouse resolution so it's not too sensitive
-        self.mouse.display_size = \
-            (self.SCREEN_WIDTH*self.sensitivity, self.SCREEN_HEIGHT*self.sensitivity)
+        self.mouse.sensitivity = 4
 
         return True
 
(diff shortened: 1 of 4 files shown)


Fruit_Jam/Larsio_Paint_Music/input_handler.py
```diff
@@ -34,182 +28,60 @@ def __init__(self, screen_width, screen_height, staff_y_start, staff_height):
...
-                self.mouse_x += x
-                self.mouse_y += y
-
-                # Ensure position stays within bounds
-                self.mouse_x = max(0, min(self.SCREEN_WIDTH - 1, self.mouse_x))
-                self.mouse_y = max(0, min(self.SCREEN_HEIGHT - 1, self.mouse_y))
-
-                return True
-
+        self.mouse = find_and_init_boot_mouse()
```
FoamyGuy, line 38: This raises an exception when the app is run without Fruit Jam OS files present
```
Error with I2S audio: [Errno 2] No such file/directory: /launcher_assets/mouse_cursor.bmp
Initializing sound manager...
Fatal error: D10 in use
```
There is fallback code here: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/Fruit_Jam/Larsio_Paint_Music/code.py#L101-L116 attempting to try again due to catching generic exceptions. But the second try also fails because on of the DAC pins is in use. 

I think we could make use of the new `deinit()` function added here: https://github.com/adafruit/Adafruit_CircuitPython_FruitJam/pull/23/files in order to resolve the trying again issue. 

As for the cursor bmp I lean towards just working around it by passing a placeholder image i.e.
```
self.mouse = find_and_init_boot_mouse(cursor_image="sprites/lars_note.bmp")
```
This app is also managing the visual cursor itself rather than using the one from the usb_host_mouse library. But in this case I think it makes more sense to have the app handle it since it's also using custom changing cursor images based on the notes selected doing so under usb_host_mouse object with the current API would be very tricky I think.

FoamyGuy (changes requested):
Need a small fix to resolve crash about missing cursor image.

Also one other request about a fix for the audio driver fallback logic which is unrelated to the main change of this PR but was uncovered while testing.

Fruit_Jam/Larsio_Paint_Music/input_handler.py
TheKitty: Just a generic note: We all made our initial programs before the USB and FruitJam libraries were fleshed out. Breakout will need some extensive rework when I get to it.

Maybe a short guide on best practices to write programs that use best practices on the Fruit Jam and FJOS?

RetiredWizard: > There is fallback code here: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/Fruit_Jam/Larsio_Paint_Music/code.py#L101-L116 attempting to try again due to catching generic exceptions. But the second try also fails because on of the DAC pins is in use.

The fallback has been a bit troubling to me, it's actually attempting to give up on the DAC and use PWM audio connected to board.D10. Of course most of the times I've encountered the fallback the issue wasn't actually that the I2S DAC wasn't available but as you encountered some other error was caught instead. In those cases falling back to PWM is never going to work. However if a speaker was connected to board.D10 and there was no DAC, I think the fallback would actually work as intended without having to add the deinit. That being said I don't see any problem adding a deinit and it's certainly cleaner. If it were my program I would probably handle the fall back differently but my goal in updating a learning guide program has been to change as little as possible to address the specific issue I'm trying to solve. If I were to be completely thorough, I probably should go back and compare the learn guide with the changes I'm making to see if any of the guide details need to be updated. but I've opted to trying to minimize the changes instead :grin:

RetiredWizard:
There is a second PWM fallback in code.py but I didn't see an easy way to run the deinit there. I'll think about it but I'm not sure it's necessary.

RetiredWizard:
We don't want to merge this until https://github.com/adafruit/Adafruit_CircuitPython_USB_Host_Mouse/pull/6 has been merged....

PR #3129, opened 2025-09-29
Minesweeper: add mouse "chording" feature

RetiredWizard:
@makermelissa This adds the mouse "chording" (both keys pressed) feature we discussed briefly on an earlier PR. 

I believe in the Microsoft version the highlighted tiles used the OPEN [ 0] sprite but I'm using the MINE_QUESTION_OPEN [15] sprite because it's not being used anywhere else and therefore I don't need to set up a separate tracking array for the highlighted cells. I could update the sprite sheet to replace the question mark with a blank open cell but I think the question mark works fine.

### DIFF
--- a/Metro/Metro_RP2350_Minesweeper/code.py
+++ b/Metro/Metro_RP2350_Minesweeper/code.py
@@ -266,9 +266,9 @@ while True:
 
         if left_button and right_button and not chord_selected:
             chord_coords = ((mouse_tg.x - ms_board.x) // 16, (mouse_tg.y - ms_board.y) // 16)
-            chord_selected = True
-            game_logic.square_chord_highlight(chord_coords)
-            waiting_for_release = True
+            chord_selected = game_logic.square_chord_highlight(chord_coords)
+            if chord_selected:
+                waiting_for_release = True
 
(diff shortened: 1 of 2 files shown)


RetiredWizard:
I added dx, dy to _get_board because pylint didn't like the number of if statements in my new methods. Calling _get_board without the dx,dy parameters results in the method behaving exactly as it did before so there should be no further changes required.

That being said, I could also update all the existing calls to _get_board to use the new parameters reducing the number of if statements in other methods. This would result in the _get_board function being called more often which probably is not quite as speed efficient. If you like this idea, I could also update _set_board in the same way although the only place it would impact is the new square_chord_highlight method.

I could also restore the _get_board function as you originally had it and just disable the pylint too many branches check.

RetiredWizard:
Out of curiosity I asked chatgpt to help with resolving the pylint too many branches issue and it came up with this approach which I liked.

While testing I also noticed that clicking outside the game board and then moving the mouse to the board without releasing the mouse button was causing a crash so I added some tests to check for out of board clicks.

makermelissa (approved):
Looks good. Tested and works great.

PR #3142, opened 2025-10-15
Logic_Gates: Fix recursion and NotImplementedErrors

RetiredWizard:
This PR primarily resolves two crashes. 

The first is a pystack memory error caused by too many recursive calls. When an entity's input was connected to its output an update would result in an infinite number of recursive calls to the value method as the update method traversed back through the input entities . This PR adds a "recursion_guard" to each instance of an entity which the value method checks and if it finds that the value method is being called a second time for the same instance the recursive loop is exited.

There was also an issue when an entity that didn't have any output value was placed next to an entity that was looking for input. An update would cause a NotImplementedError when the value method was called on the input only entity (OutputPanel & NeoPixelOutput).

I also slightly tweeked the mouse parsing loop as I was running into an issue with the first toolbox item being selected erroneously if the mouse was being moved as the toolbox was selected. Waiting for the mouse button to be released before processing the click also opens the possibility of supporting a drag feature which I'm thinking might allow for a nice future enhancement for dragging wire connections.

With these changes in I was actually able to model an S/R latch. Even with the recursion guard I did have to increase the CIRCUITPY_PYSTACK_SIZE value, I went to 8000 but I didn't do any experimenting to see if a smaller value was sufficient.

### DIFF
--- a/Fruit_Jam/Fruit_Jam_Logic_Gates/workspace.py
+++ b/Fruit_Jam/Fruit_Jam_Logic_Gates/workspace.py
@@ -407,20 +407,6 @@ class Workspace:
             if self.moving_entity is not None:
                 self.moving_entity = None
                 self.mouse_moving_tg.hidden = True
-            else:
-                # apply offset value based on scroll position
-                mouse_x -= self._scroll_x * self.tilegrid.tile_width * 1
-                mouse_y -= self._scroll_y * self.tilegrid.tile_height * 1
-
-                # calculate tile coordinates
-                tile_x = mouse_x // self.tilegrid.tile_width
-                tile_y = mouse_y // self.tilegrid.tile_height
-
-                # get the entity at the coordinates if there is one
-                hovered_entity = self.entity_at((tile_x, tile_y))
-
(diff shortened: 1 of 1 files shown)


RetiredWizard:
If you think you want to eventually implement the enhancements you described, I'll leave this as is. However, if you're okay with me tweaking it some more I have some ideas. 

I think the program is great now as a simple logic demonstrator but it struggles if you try and use it as a digital circuit simulator. The first thing I was thinking about was the wire implementation, I'm thinking that multiple connected wires should bypass all the intermediate tiles and get the output directly from the first input wire. I haven't really worked out the details of how to implement that yet but I noticed as I built a larger circuit the updates started taking a long time to propagate across the multi-segment wires.

I haven't dug into the code yet, but just by using the application it looks to me like the wires may get confused as to where they're picking up input from when there are multiple entities adjacent so that's something I'd probably also like to dig into further.

Once wires are working more efficiently, I'd like to tackle some of the more subtle circuit simulation issues and then probably the graphics improvements you mentioned.

PR #3145, opened 2025-10-22
Clarify leading zero handling in time display

TheKitty:
Per guide feedback where a user wanted suppressed zero on times from 01:00 to 09:59. Guide will be augmented per Ladyada to allow the user to select which behavior they want just by commenting out a line and uncommenting out the desired behavior Default is the code as written that does not suppress a zero.


BlitzCityDIY (approved):
looks good!

PR #3167, opened 2025-11-18
resolve mouse errors

BlitzCityDIY:
hihi @FoamyGuy - i kept getting a mouse read error in the memory game. this seems to have resolved it (based on the mouse usb host code). wanted you to look at it first before i took it out of draft in case there was context i wasn't aware of

### DIFF
--- a/Metro/Metro_RP2350_Memory/memory_game/code.py
+++ b/Metro/Metro_RP2350_Memory/memory_game/code.py
@@ -10,6 +10,7 @@ Players trade off using the USB mouse to play their turns.
 import array
 import random
 import time
+import atexit
 from displayio import Group, OnDiskBitmap, TileGrid
 from adafruit_display_text.bitmap_label import Label
 from adafruit_display_text.text_box import TextBox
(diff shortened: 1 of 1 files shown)


FoamyGuy:
I tested this and confirmed that without the `atexit` code the keyboard (of a keyboard/mouse combo) no longer works in the Fruit Jam OS launcher if you launch memory play a game and then click exit at the end. 

I believe the re-attach thing that it's doing in the atexit call gets it back so the keyboard is hooked up as stdin.

I will do some more testing with the mouse from the store to see if I can reproduce the read errors with that and try to resolve them. 

Or maybe see if the launcher can detect no keyboard attached to kernel and have the code in there to re-attach it rather than `atexit` inside of the apps where it got un-attached.

FoamyGuy (approved):
Looks good to me, thank you!

If we do see UsbError pop-up from the code in atexit with the basic mouse, I think it may be okay to just catch that exception and pass. The mouse shouldn't need to re-attach so I think everything else may still work if we catch that exception and let it go on.

RetiredWizard:
Just a late thought... There were some relatively recent changes to the core USB support that resulted in some of the device attributes frequently throwing USBError (device.idVendor, device.idProduct, device.manufacturer, etc).

If you're still seeing errors, you might try catching that error with a try/except/pass around the information print statements.

PR #3169, opened 2025-11-24
FruitJam PyPaint and Memory: Add support for Composite HID devices

RetiredWizard:
This is one of three library changes (Adafruit_USB_Host_Descriptors and Adafruit_CircuitPython_USB_Host_Mouse) that provides a framework to support combination keyboard/trackpad USB devices.

I don't believe CircuitPython can have both the keyboard and trackpad of a combination device active at the same time but being able to have an application decide which one is active would be useful.

These changes to the PyPaint program allow it to search for non-boot (report) mice if no boot mice are found. An atexit callback is also added to switch control back to the keyboard when a non-boot mouse was configured for use with the application.

The trackpad sensitivity was also increased to make it more usable.

### DIFF
--- a/Metro/Metro_RP2350_Memory/memory_game/code.py
+++ b/Metro/Metro_RP2350_Memory/memory_game/code.py
@@ -11,7 +11,7 @@ import sys
 import random
 import time
 import atexit
-from displayio import Group, OnDiskBitmap, TileGrid
+from displayio import Group, OnDiskBitmap, TileGrid, CIRCUITPYTHON_TERMINAL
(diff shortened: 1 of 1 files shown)


RetiredWizard:
I decided to try refactoring the memory game for the library changes as well. Obviously this PR depends on the two USB PRs being merged first (https://github.com/adafruit/Adafruit_CircuitPython_USB_Host_Mouse/pull/14 and https://github.com/adafruit/Adafruit_CircuitPython_USB_Host_Descriptors/pull/9)

PR #3173, opened 2025-12-02
Metro2350_Match3: Add support for HID multi device keyboards/trackpads

RetiredWizard:
This PR adds support for keyboards with integrated mice/trackpads. CircuitPython can't use more than on device on a combination keyboard/pointing device at the same time, but this update allows the pointing device to be used during the game and then reattached the keyboard device when the game exits.

This PR depends on a pending PR to the adafruit_usb_host_descriptors library so I'm submitting as a draft until the PR gets merged.

### DIFF
--- a/Metro/Metro_RP2350_Match3/match3_game/code.py
+++ b/Metro/Metro_RP2350_Match3/match3_game/code.py
@@ -305,9 +305,6 @@ for find_endpoint, default_sync in [
 
                 detached_interfaces.append(detached)
 
-                # set the mouse configuration so it can be used
-                #device.set_configuration()
-
         except usb.core.USBError as e:
             # The mouse might have glitched and may not be detected but at least we don't crash
             print(e)
@@ -318,6 +315,7 @@ for find_endpoint, default_sync in [
     if len(mice) >= 2:
         break
 
+# set the mouse configuration on any detected mice so they can be used
 for device in mice:
     device.set_configuration()
 


RetiredWizard:
Since the Adafruit_CircuitPython_USB_Host_Mouse library does not yet support returning multiple devices, this game doesn't use the library and only depends on the adafruit_usb_host_descriptors library for which the required PR has been merged (although I don't think the update is published to CIRCUP yet).

I'm marking this as ready to merge.

Metro/Metro_RP2350_Match3/match3_game/code.py
```diff
@@ -445,6 +490,11 @@ def atexit_callback():
                 # get the current mouse coordinates
                 coords = (mouse_tg.x, mouse_tg.y, 0)
 
+                # if the mouse point is within the exit
+                # button bounding box
+                if exit_btn.contains(coords):
+                    supervisor.reload()
```
FoamyGuy: This was moved out of the `else` block for `if match3_game.cur_state != STATE_GAMEOVER:` just below here. 

Previously being in that `else` block meant that clicking within the bounds of the exit_btn would only be registered if it was in the GAMEOVER state. It looks to me like now it checks even when it's not GAMEOVER, but the exit button isn't visible so I think this could cause an invisible click area to run the `supervisor.reload()` if it were clicked.

Was there a reason for moving this out of the `else` block? And is there something else I haven't noticed that limits it to only take effect when the game is actually over?

RetiredWizard: I moved the exit button to the edge of the display while the game was not in the GAMEOVER state so that the game could be exited using the mouse (the keyboard is disabled if all you have is the combo device attached). When the game is over the exit button is moved back to the location on top of the gameover displayio group.

PR #3236, opened 2026-04-26
update actions to Node.24; remove unneeded `if:` from arduino_cli.yml

### LINKED ISSUE #3234: Fix arduino_cron.yml
Warning from CI runs:
>Workflow syntax warning: .github/workflows/arduino_cron.yml#L32
.github/workflows/arduino_cron.yml (Line: 32, Col: 13): Conditional expression contains literal text outside replacement tokens. This will cause the expression to always evaluate to truthy. Did you mean to put the entire expression inside ${{ }}?

### LINKED ISSUE #3235: Update to Node.24 actions
> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/checkout@v4, actions/setup-python@v5, actions/upload-artifact@v4, softprops/action-gh-release@v1.

dhalbert:
- Fixes #3235.

Update CI actions to their latest versions, mainly to use Node.24

- Fixes #3234.

This `if:` was always true:
```
        if: ${{ steps.check-cron.outputs.status }} == false && ${{ steps.check-dispatch.outputs.status }} == false
```
I first removed the curly brace expansions, but that caused it to be false sometimes, which caused the `is-needed` step to fail, because because `steps.check-updated.outputs.status` was not set. But `is-needed` was already checking the values in the `if:`. So the `if:` wasn't really needed (it could save a little time, but not a big deal).

### DIFF
--- a/.github/workflows/arduino_cron.yml
+++ b/.github/workflows/arduino_cron.yml
@@ -12,7 +12,7 @@ jobs:
     outputs:
       answer: ${{ steps.is-needed.outputs.answer }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
         with:
           fetch-depth: 2
       - name: Check if run by adabot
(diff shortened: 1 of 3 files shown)


brentru (approved):
Agree that the if is unnecessary, LGTM

PR #3239, opened 2026-05-07
CLUE ble beacon remote

videopixil:
CLUE ble remote code for magic devices

### DIFF
--- a/CLUE_BLE_Beacon_Remote/code.py
+++ b/CLUE_BLE_Beacon_Remote/code.py
@@ -1,0 +1,693 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
+#
+# SPDX-License-Identifier: MIT
+'''MagicBand+ BLE remote for the Adafruit CLUE (nRF52840).
+
+Broadcasts Disney MagicBand+ BLE commands using the 0x0183 manufacturer
+identifier. Grid of category tiles on startup; A/B navigate, double-tap B
+opens the selected category. In the list view, A/B scroll, double-tap A
+fires the highlighted command, double-tap B returns to grid, and a long
+press on B fires the OFF command to cancel a running animation. Shake the
+CLUE to pick a random command; confirm with double-tap A or cancel with B.
+'''
+# Target: Adafruit CLUE (nRF52840) - the BLE remote
+import gc
+import os
+import random
+import time
+
+import _bleio
+import alarm
+import board
+import digitalio
+import microcontroller
+import neopixel
+import pwmio
+import supervisor
+from adafruit_debouncer import Button
+
+import ble_transmitter
+import command_library
+import ui
+
+_STATE_GRID = 0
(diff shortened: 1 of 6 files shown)


videopixil:
@TheKitty @BlitzCityDIY CLUE ble remote code is ready to checkout

CLUE_BLE_Beacon_Remote/code.py
```diff
@@ -0,0 +1,693 @@
...
+_pixel = neopixel.NeoPixel(board.NEOPIXEL, 1, brightness=0.3)
+_pixel.fill((0, 0, 0))
+
+# PWM-driven speaker avoids per-note audio buffer allocations.
+_speaker = pwmio.PWMOut(board.SPEAKER, variable_frequency=True, duty_cycle=0)
+
+_btn_a_io = digitalio.DigitalInOut(board.BUTTON_A)
+_btn_a_io.switch_to_input(pull=digitalio.Pull.UP)
+_btn_b_io = digitalio.DigitalInOut(board.BUTTON_B)
+_btn_b_io.switch_to_input(pull=digitalio.Pull.UP)
```
BlitzCityDIY: none of these should be internal, they're being called directly

CLUE_BLE_Beacon_Remote/code.py
```diff
@@ -0,0 +1,693 @@
...
+# ValueError - that's fine, we fall back to display.brightness only.
+_backlight_pin = None
+try:
+    _backlight_pin = digitalio.DigitalInOut(board.TFT_LITE)
+    _backlight_pin.switch_to_output(value=True)
+except (AttributeError, ValueError, RuntimeError):
+    # Display module has claimed TFT_LITE - fine, we'll use
+    # display.brightness alone to control the backlight.
+    pass
+
```
BlitzCityDIY: if the `display.brightness` is used and its known that the TFT_LITE pin is in use then this can be taken out

CLUE_BLE_Beacon_Remote/code.py
```diff
@@ -0,0 +1,693 @@
...
+        # Filesystem is read-only - host has ownership. Set NVM flag
+        # so next reset auto-creates the marker and enters capture mode.
+        try:
+            microcontroller.nvm[0] = 1
+            listen_view.set_status("Reset to enable save  B", 0xFF8000)
+        except (AttributeError, ImportError):
+            listen_view.set_status(f"Save fail: {err}  B", 0xFF0000)
+
+
+def _wait_for_dismiss_press():
```
BlitzCityDIY, line 222: this feels over complicated, especially with the time.sleep calls for a debounce while using the debounce library

BlitzCityDIY (changes requested):
hihi- just some minor fixes, mainly formatting

videopixil:
sweet thanks @BlitzCityDIY updated code.py with recommend formatting, tested on hardware and all works great

PR #3240, opened 2026-05-07
BLE beacon ears

videopixil:
QTPy code to make into a BLE receiver

### DIFF
--- a/BLE_Beacon_Ears/code.py
+++ b/BLE_Beacon_Ears/code.py
@@ -7,10 +7,12 @@ Scans for Disney BLE adverts and renders matching commands on stereo
 NeoPixel Jewels. Two logical zones (left ear, right ear) are rendered
 with stereo phase offsets to give static colors a gentle out-of-phase
 breathing animation and rotations a left-leads-right sweep.
-
-To switch from real Jewels to onboard NeoPixel for testing without the
-ear hardware, flip _USE_ONBOARD to True below.
 '''
+# pylint: disable=redefined-outer-name
+# Setup and main loop are inlined at module level (no main() wrapper).
+# Helper functions defined above take parameters named the same as the
+# inlined module-level loop variables (zone, t, etc) - this is expected
+# and harmless: helpers and the inline body never share scope at runtime.
 # Target: Adafruit QT Py ESP32-S3 - the BLE Beacon Ears
 import math
 import random
@@ -25,9 +27,6 @@ import battery as battery_mod
 import magicband_protocol
 import pixel_zones
 import renderer
-
-# Flip this to True to test on the QT Py's onboard pixel without jewels.
-_USE_ONBOARD = False
 
 # 15 fps target -> ~66ms per frame.
 _TARGET_FPS = 15
@@ -66,7 +65,7 @@ _BRIGHTNESS_PRESETS = (0.02, 0.04, 0.08)
 # How long the BOOT button must be held to count as a long press.
 # Short press cycles brightness preset (the more frequently-used action).
 # Long press shows the battery level (less frequent, requires intent).
-_LONG_PRESS_S = 0.5
+LONG_PRESS_S = 0.5
 _BRIGHTNESS_FLASH_DURATION_S = 0.6
 
 # USB-presence detection threshold (raw QT Py voltage reading).
@@ -146,8 +145,8 @@ _SOLO_EXIT_RGB = (40, 100, 220)
 # committing a multi-press count. 350ms is a common sweet spot - tight
 # enough that single-press brightness cycling still feels responsive,
 # wide enough that comfortable triple-presses register reliably.
-_BUTTON_SHORT_MS = 350
-_BUTTON_LONG_MS = int(_LONG_PRESS_S * 1000)
+BUTTON_SHORT_MS = 350
+BUTTON_LONG_MS = int(LONG_PRESS_S * 1000)
 
 
 def _build_showpieces():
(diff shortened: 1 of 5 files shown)


videopixil:
The receiver code.py has lint disables on main() for
too-many-locals, too-many-branches, and too-many-statements.
These are intentional and supported by direct on-hardware bisection.
Symptom: Refactoring main() into additional helper functions
causes import _bleio to fail at boot with
espidf.IDFError: Generic Failure, even after USB power-cycle.
Standalone _bleio import in the REPL works fine, the platform is
healthy, and gc.mem_free() reports 120KB+ free with 20KB
contiguous allocation succeeding - so it is not memory pressure
in the way that's typically reported.

Implication for contributors: if you extend code.py and
adding code pushes the file over ~32.5KB, the firmware will fail
to boot. The remote-trigger if/elif cascade in the BLE scan loop
is intentionally inline for this reason (extracting it to helpers
puts us over the cliff). When in doubt, prefer adding code to the
helper modules (renderer.py, magicband_protocol.py) which
don't have this constraint.

videopixil:
@TheKitty @BlitzCityDIY QTPy ble receiver code is ready to checkout - lint disables on main() for
too-many-locals, too-many-branches, and too-many-statements explained above

TheKitty (approved):

BLE_Beacon_Ears/code.py
```diff
@@ -0,0 +1,791 @@
...
+        _render_battery_display(zones, state.battery_display_voltage, t)
+    elif state.active_state is not None:
+        _render_active_state(zones, state, frame_t)
+    elif state.solo_mode:
+        _render_solo_cycle(zones, state, frame_t)
+    else:
+        renderer.render_idle(zones)
+
+
+def main():
```
BlitzCityDIY: `code.py` usually does not utilize a `main()`. the code segment before `while True:` could go at the top of the code and then the `while True` loop could be at the bottom of the code

BLE_Beacon_Ears/code.py
```diff
@@ -0,0 +1,791 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
```
BlitzCityDIY, line 1: just a general question- is all of this code being used in the guide? i know there were multiple apps being prototyped so just want to check that the beacon ears need all of this code

BLE_Beacon_Ears/code.py
```diff
@@ -0,0 +1,791 @@
...
+import math
+import random
+import time
+
+import _bleio
+import board
+import digitalio
+from adafruit_debouncer import Button
+
+import battery as battery_mod
```
BlitzCityDIY, line 26: what is this battery import? it is not in any of the helper files and when i try to run the code on a qt py esp32-s3 i get an `ImportError`:
```
  File "code.py", line 24, in <module>
ImportError: no module named 'battery'
```

BLE_Beacon_Ears/code.py
```diff
@@ -0,0 +1,791 @@
...
+
+    batt = battery_mod.BatteryMonitor()
+
+    button_pin = digitalio.DigitalInOut(board.BUTTON)
+    button_pin.switch_to_input(pull=digitalio.Pull.UP)
+    button = Button(button_pin, value_when_pressed=False,
+                    short_duration_ms=_BUTTON_SHORT_MS,
+                    long_duration_ms=_BUTTON_LONG_MS)
+
+    state = _RuntimeState()
```
BlitzCityDIY: the purpose of a leading underscore in python is to mark code as internal use only, meaning that the end user wouldn't access it directly. this file uses it a lot for variables and methods that are being called directly in user facing code

BLE_Beacon_Ears/code.py
```diff
@@ -0,0 +1,791 @@
...
+    return bytes(payload[:3]) == _REMOTE_BRIGHTNESS_PACKET
+
+
+def _is_remote_find_trigger(payload):
+    return bytes(payload[:3]) == _REMOTE_FIND_PACKET
+
+
+def _is_remote_statue_trigger(payload):
+    return bytes(payload[:3]) == _REMOTE_STATUE_PACKET
+
```
BlitzCityDIY: i think this could be condensed into one lookup function that checks the possible commands:

```
REMOTE_COMMANDS = {
    bytes.fromhex("aa4201"): "battery",
    bytes.fromhex("aa4203"): "brightness",
    bytes.fromhex("aa4204"): "find",
    bytes.fromhex("aa4205"): "statue",
}

def remote_command(payload):
    return REMOTE_COMMANDS.get(bytes(payload[:3]))
```

BLE_Beacon_Ears/code.py
```diff
@@ -0,0 +1,791 @@
...
+                    print("[statue preview] firing golden swirl")
+                    fake_statue_cmd = {
+                        "kind": "statue_beacon",
+                        "statue_id": "PV",
+                        "raw": bytes(payload),
+                    }
+                    candidate = renderer.for_command(fake_statue_cmd)
+                    if candidate is not None:
+                        new_command = (candidate, entry.rssi, bytes(payload))
+                    continue
```
BlitzCityDIY: and then you can call the command and perform logic depending on the result:

```
command = remote_command(payload)
if command is None:
    continue

now = time.monotonic()
if now - state.last_trigger_time < TRIGGER_COOLDOWN_S:
    continue
state.last_trigger_time = now

if command == "brightness":
    # etc
```

BLE_Beacon_Ears/pixel_zones.py
```diff
@@ -0,0 +1,161 @@
...
+end. Double-buffering is not needed at 15fps - a torn frame would be
+invisible to the eye.
+'''
+# Target: Adafruit QT Py ESP32-S3 - the BLE Beacon Ears
+import board
+import digitalio
+import neopixel
+
+
+class OnboardSingle:
```
BlitzCityDIY: i would take this out since it was used for prototyping only and just uses the onboard neopixel

BLE_Beacon_Ears/pixel_zones.py
```diff
@@ -0,0 +1,161 @@
...
+            self._left.show()
+            self._right.show()
+            self._last_shown_black = True
+        else:
+            self._left.show()
+            self._right.show()
+            self._last_shown_black = False
+
+
+class _SingleZone:
```
BlitzCityDIY: seems like it is part of the prototyping, could remove

BlitzCityDIY (changes requested):
okay! here's a summary of the comments i'd like you to look at:

1. most important is the `import battery`. it's throwing an import error, not sure if there is another file that got missed?
2. consider not using a `main()`. the code setup should be able to happen before the loop and then go into the loop
3. consider not using internal declarations for so many of the variables, methods and functions that are then called in `code.py`
4. the checks for the incoming commands could be condensed
5. anything used for prototyping could be removed. that stuff could confuse folks if they're following along with the guide or trying to edit the code

good to see the debouncer library being used, i think it helps to simplify the button press tracking

BLE_Beacon_Ears/code.py
BlitzCityDIY: just saw the other open PR, got it

videopixil:
@BlitzCityDIY awesome thanks! applied fixes with all of the suggested sections and added the battery.py file. verified new files all still work with hardware - stating on the CLUE edits next

PR #3252, opened 2026-06-24
Fix for hold-to-save

videopixil:
Fix hold-to-save: time the hold from the debounced level, not edge flags

The Button.pressed/released one-shot edge flags intermittently dropped a press, so hold-to-save would fall through and register as a short tap (switching mode instead of saving). poll_button now reads the debounced level (button.value) and detects the press/release edges itself, timing the hold with time.monotonic(). This also survives the ~300-500 ms loop stall from the blocking 1 Hz GPS read, which the library's internal hold timer could not. The debouncer is still used for noise filtering.


videopixil:
@BlitzCityDIY @TheKitty Pushed a fix for hold-to-save. While demoing I found the button hold had stopped saving waypoints; a quick tap would switch mode, but a long hold also just switched mode instead of opening the name picker.

Minecraft_Compass/code.py
BlitzCityDIY, line 1: can you revert code.py and mc_config.py? it looks like no code changed, just file mode

BlitzCityDIY (changes requested):

BlitzCityDIY (approved):

PR #3253, opened 2026-06-24
CPB Morse code paddle project code

FoamyGuy:
both boot.py and code.py required for this project.

Morse paddle can be used for live keying and CPB built-in buttons can be pre-programmed with automatic messages.

### DIFF
--- a/CPB_Morse_Code_Paddle/code.py
+++ b/CPB_Morse_Code_Paddle/code.py
@@ -1,0 +1,167 @@
+# SPDX-FileCopyrightText: 2026 Tim Cocks for Adafruit Industries
+#
+# SPDX-License-Identifier: MIT
+"""USB audio Morse code paddle. Connect Morse paddle 3.5mm jack to pin A1 and GND.
+
+Connect the board to a host computer and select the CircuitPython USB
+microphone as a recording/input device.
+
+Use the paddle to enter Morse code live, update the button messages to
+automatically convert and send pre-programmed strings with the built-in
+buttons.
+"""
+
+import time
+
+import board
+import digitalio
+import synthio
+import usb_audio
+import adafruit_debouncer
+
+# Configuration
+
+BTN_A_MESSAGE = "HELLO WORLD"
+BTN_B_MESSAGE = "CIRCUITPYTHON"
+
+TONE_NOTE = 60  # synthio note number for the tone pitch (MIDI note, 60 = C4)
+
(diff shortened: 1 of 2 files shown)


CPB_Morse_Code_Paddle/code.py
```diff
@@ -0,0 +1,167 @@
...
+
+    if morse_paddle_debounce.fell:
+        synth.press(60)
+    if morse_paddle_debounce.rose:
+        synth.release(60)
+
+    if btn_a_debounce.rose:
+        play_morse(BTN_A_MESSAGE)
+    if btn_b_debounce.rose:
+        play_morse(BTN_B_MESSAGE)
```
dhalbert: I would prefer if you used `keypad` here. It's simpler. Also right now `debouncer` triggers at the end of a bounce period, which increases latency. (I want to fix that.)

TheKitty (approved):

CPB_Morse_Code_Paddle/code.py
```diff
@@ -0,0 +1,167 @@
...
+btn_a.switch_to_input(digitalio.Pull.DOWN)
+btn_a_debounce = adafruit_debouncer.Debouncer(btn_a)
+
+btn_b = digitalio.DigitalInOut(board.BUTTON_B)
+btn_b.switch_to_input(digitalio.Pull.DOWN)
+btn_b_debounce = adafruit_debouncer.Debouncer(btn_b)
+
+morse_paddle_dio = digitalio.DigitalInOut(board.A1)
+morse_paddle_dio.switch_to_input(digitalio.Pull.UP)
+morse_paddle_debounce = adafruit_debouncer.Debouncer(morse_paddle_dio)
```
dhalbert: A single-button Morse code switch is usually called a "key".
Image

"Paddles" usually means two switches, set up with one doing dits (dots) and the other dahs (dashes). They physically resemble paddles:

Image

dhalbert (changes requested):
nice!

CPB_Morse_Code_Paddle/code.py
FoamyGuy: Ahhh, I saw paddles coming up while looking around and mistakenly thought those still had a single switch but was just mounted sideways. I will adjust the references in code and guide to call it a key.

I will also switch over to keypad, I forget about it when I need debounced buttons. I like that it removes the only library requirement also.

dhalbert (approved):
thanks!

PR #3259, opened 2026-07-14
Moon_Flight_Tracker

videopixil:
LUNA TRANSIT flight tracker code.
Alerts when a plane is about to cross (or pass near) the moon, for photographing transits.
Adafruit Qualia ESP32-S3 + 2.1" round 480x480 touch TFT (product 5792)


TheKitty (approved):

TheKitty:
See https://github.com/adafruit/Adafruit_Learning_System_Guides/actions/runs/29341732969/job/87114948956?pr=3259 for issues

videopixil:
hi @BlitzCityDIY @TheKitty code is ready to checkout - let me know if fable missed anything

Moon_Flight_Tracker/code.py
```diff
@@ -0,0 +1,985 @@
...
+        g.append(text(loc_msg[0], CX, 382, loc_msg[1], 1, (0.5, 0)))
+    g.append(nav_group("LOC"))
+    return g
+
+
+BUILDERS = {"radar": build_radar, "list": build_list, "moon": build_moon,
+            "loc": build_loc, "detail": build_detail}
+
+# ---------------- touch init ----------------
+class RawCST8XX:
```
BlitzCityDIY: is this a custom class for the touch driver to get around a bug in the library? if so, an issue should be filed in the library instead to fix it.

Moon_Flight_Tracker/settings.toml.example
```diff
@@ -0,0 +1,10 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
```
BlitzCityDIY: this file can be removed. it will be confusing to include it in the project bundle. instead in the learn guide note that people need to include their settings.toml file like in other wifi projects, especially where its just ssid info

BlitzCityDIY (changes requested):

videopixil:
awesome thanks @BlitzCityDIY 

touch driver issue filed to: https://github.com/adafruit/Adafruit_CircuitPython_CST8XX/issues/4

added raw-poller-only init with the explanatory comment, the dump_id_regs() diagnostic, and the debug hooks). 

DEBUG_TOUCH = False at the top.

The dump_id_regs() method can stay — it's gated behind DEBUG_TOUCH and is useful documentation for anyone else hitting this panel batch.

code.py updated to 966 lines and tl021wvc02.py holds the display's initialization sequence — the ~55-line block of raw bytes that gets sent to the round TFT panel at boot to configure it (gamma curves, voltage settings, pixel format, sleep-out/display-on commands). TL021WVC02 is the panel model used in the Qualia 2.1" round display. Without tl021wvc02.py, code.py crashes at boot with ImportError: no module named tl021wvc02 — the display never initializes. The README copy list reflects this.

Moon_Flight_Tracker/code.py
```diff
@@ -0,0 +1,965 @@
...
+        touch.disable_auto_sleep()
+        print("touch: auto-sleep disabled")
+    except Exception as e:
+        print("touch: could not disable auto-sleep:", e)
+    print("touch: raw CST8XX @0x15 ok")
+    if DEBUG_TOUCH:
+        print("touch id regs (boot):", touch.dump_id_regs())
+except Exception as e:
+    print("touch init failed:", e)
+    touch = None
```
BlitzCityDIY: since the PR was merged for the touch driver, please remove all of this and use the proper driver init. we'll do a release and the bundle will update overnight with the latest version

Moon_Flight_Tracker/tl021wvc02.py
```diff
@@ -0,0 +1,55 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
```
BlitzCityDIY, line 1: did you try using the [qualia helper](https://github.com/adafruit/Adafruit_CircuitPython_Qualia)? it makes it so that the init is a lot simpler and you don't need to include the sequence

Moon_Flight_Tracker/code.py
```diff
@@ -0,0 +1,965 @@
...
+        poll_touch()
+
+    if screen == "loc" and loc_msg and now - loc_msg_t >= 4:
+        loc_msg = None
+        show()
+
+    # tap detection lives in poll_touch() — see notes there
+    poll_touch()
+
+    time.sleep(0.01)
```
BlitzCityDIY: where you're using `time.monotonic()` do you want this blocking delay?

BlitzCityDIY (changes requested):

videopixil:
@BlitzCityDIY okay sweet all three review items are addressed in hardware/code.py and README.md:

Touch driver — RawCST8XX (~75 lines) deleted, replaced with the stock init: touch = adafruit_cst8xx.Adafruit_CST8XX(board.I2C()). It's a drop-in — the raw poller mimicked the library's .touches API, so poll_touch() is unchanged. Dropped the now-unused I2CDevice import and the diagnostic register dumps. A comment notes it needs the bundle release with the 0x00 chip-ID fix.

Blocking delay — removed the time.sleep(0.01) at the loop tail; the loop is already paced by the touch I2C transaction, so the sleep was just added tap latency. (The only remaining time.sleep is the NTP retry backoff, which is intentionally blocking.)

Qualia helper — kept direct init, with the reason now documented in the code and README: the helper's Round21 class bakes the stock 16 MHz pixel clock into the display definition with no override parameter, and this project runs 8 MHz to suppress the WiFi-burst jitter we chased earlier. 

If you'd accept a timings/frequency override in adafruit_qualia, we could switch and drop tl021wvc02.py entirely.

BlitzCityDIY (approved):
thanks!

PR #3264, opened 2026-07-17
update ci test for ybox3

BlitzCityDIY:
new philhower release (5.7.0) adds a check that any sketch with OTA has a FS defined. changing the check to use tinyusb pico w for ybox3. tested locally


BlitzCityDIY:
looking into the error for the lumon speaker code

TheKitty (approved):

PR #3275, opened 2026-08-06
Not_A_Typewriter USB host: retry keyboard enumeration

noise:
In testing with my build, the pi was failing to find the USB keyboard 4/5 boots. PR adds a retry of usb.core.find() for ~10s so the keyboard has time to appear. In my debugging runs it typically took 2 retries but likely varies with the kb hardware


BlitzCityDIY (approved):

BlitzCityDIY:
thanks!

PR #3287, opened 2026-09-09
eink message board

videopixil:
Off-grid LoRa mesh message board on a 4.2" tri-color e-ink panel.
Receives and decodes messages compatible with the Meshtastic protocol,
and shows a weather forecast, air quality and battery runtime.

Receive only. This board does not transmit and will not appear in the
mesh node list.


videopixil:
@TheKitty @BlitzCityDIY I think this one is ready to review. I don't have permissions to push the library, let me know how I could send that in to review

TheKitty (commented):
Youe settings.toml example seems to have your real data?

videopixil:
@TheKitty okay removed info thanks for checking!

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+    "nodes": "Nodes",
+    "weather": "Weather",
+}
+
+
+
+
+
+
+
```
BlitzCityDIY: can remove these empty lines

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+
+
+
```
BlitzCityDIY: remove these empty lines

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+    if weather is not None and elapsed - weather_fetched < WEATHER_INTERVAL:
+        return
+    fetched = fetch_weather(USE_FAHRENHEIT)
+    if fetched is not None:
+        weather = fetched
+        weather_fetched = elapsed
+
+
+
+
```
BlitzCityDIY, line 143: remove these empty lines

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+try:
+    microcontroller.cpu.frequency = CPU_FREQUENCY
+    print("cpu at", microcontroller.cpu.frequency // 1_000_000, "MHz")
+except (AttributeError, ValueError, RuntimeError) as clock_error:
+    print("cpu frequency unchanged:", clock_error)
+
+displayio.release_displays()
```
BlitzCityDIY: i would call this first in the code, usually its called right after the imports

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+rfm9x._write_u8(  # pylint: disable=protected-access
+    SYNC_WORD_REG, MESHTASTIC_SYNC_WORD
+)
+
+# Do not claim NEOPIXEL_POWER or I2C_POWER here. On this board
+# board.STEMMA_I2C() already manages the accessory rail, and
+# driving it again knocks the STEMMA sensor off the bus.
+
+try:
+    import neopixel
```
BlitzCityDIY: please have all imports at the top of the code. don't wrap it in a try/except. was there a purpose for using a try/except here?

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+
+
+button = digitalio.DigitalInOut(BUTTON_PIN)
+button.direction = digitalio.Direction.INPUT
+button.pull = digitalio.Pull.UP
+
+
+
+
+
```
BlitzCityDIY, line 242: remove these empty lines

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+if not all(mesh_fonts.FONTS.values()):
+    print("bitmap fonts not found, falling back to terminalio")
+
+
+
+
+
+
+
+
```
BlitzCityDIY, line 273: remove these empty lines

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+
+
```
BlitzCityDIY: remove these empty lines

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+            text=right,
+            color=BLACK,
+            scale=status_scale,
+            anchor_point=(1.0, 0.5),
+            anchored_position=(WIDTH - MARGIN, baseline),
+        )
+    )
+
+
+def draw_board():
```
BlitzCityDIY, line 658: i would have this just called directly in the loop instead of wrapping it in a function, especially where you declare these globals and they're being referenced by the text graphics. for example, i see `air` is referenced in `draw_header()`. if you need to update text, you can update the text property for the object with the reading from the sensor. ex: `air_text.text = air`

MeshFruit_EInk_Message_Board/code.py
```diff
@@ -0,0 +1,860 @@
...
+    # Active low with an internal pull up. The radio call above blocks
+    # for up to a second, so a press can take that long to register.
+    now = time.monotonic()
+    if not button.value:
+        if press_started is None:
+            press_started = now
+    elif press_started is not None:
+        held = now - press_started
+        press_started = None
+        if held > DEBOUNCE_SECONDS:
```
BlitzCityDIY: in previous projects i've mentioned using the Debouncer library or keypad for managing debouncing and long press/short press depending on needs. please look at either of those for this as well.

MeshFruit_EInk_Message_Board/mesh_fonts.py
```diff
@@ -0,0 +1,94 @@
...
+    try:
+        return bitmap_font.load_font(path)
+    except (OSError, ValueError) as error:
+        print("font", path, "unavailable:", error)
+        return None
+
+
+FONTS = {name: load_font(path) for name, path in FONT_PATHS.items()}
+
+# Fall back to the built-in face so the board still runs with no font
```
BlitzCityDIY: where folks will be installing the project bundle can you assume that they will have the fonts and instead of fallback code to terminalio you could always have it print that its missing the fonts which will let folks know they need to add them

MeshFruit_EInk_Message_Board/mesh_fonts.py
```diff
@@ -0,0 +1,94 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
```
BlitzCityDIY: i wonder if this code could just be included in code.py?

MeshFruit_EInk_Message_Board/mesh_icons.py
```diff
@@ -0,0 +1,164 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
+#
+# SPDX-License-Identifier: MIT
+
+"""Weather glyphs drawn with vectorio, so no bitmap assets are needed."""
```
BlitzCityDIY, line 5: did you hit `MemoryError`'s with bitmaps? we have a lot of really nice weather graphics so i'm just curious on the history here

MeshFruit_EInk_Message_Board/mesh_sensors.py
```diff
@@ -0,0 +1,185 @@
...
+
+AIR_READ_ATTEMPTS = 3
+
+
+AIR_READ_RETRY_SECONDS = 1.0
+
+
+AIR_SETTLE_SECONDS = 1.0
+
+# Below this percent-per-hour the gauge's rate is noise, so no
```
BlitzCityDIY: remove this repeat comment, appears again directly below

MeshFruit_EInk_Message_Board/mesh_sensors.py
```diff
@@ -0,0 +1,185 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
+#
+# SPDX-License-Identifier: MIT
+
+"""Air quality sensor and battery gauge.
+
+Bring these up before the display, radio or NeoPixel.
+board.STEMMA_I2C() claims and manages I2C_POWER itself, and if
```
BlitzCityDIY: does this comment mean that you were having power dips?

MeshFruit_EInk_Message_Board/mesh_sensors.py
```diff
@@ -0,0 +1,185 @@
...
+
+# Below this percent-per-hour the gauge's rate is noise, so no
+# runtime estimate is shown.
+
+
+# Below this percent-per-hour the gauge's rate is noise, so no
+# runtime estimate is shown.
+BATTERY_RATE_FLOOR = 0.5
+
+# Bring the STEMMA sensor up before anything else touches hardware.
```
BlitzCityDIY: remove this repeat comment, appears again directly below

MeshFruit_EInk_Message_Board/mesh_sensors.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+# Bring the STEMMA sensor up before anything else touches hardware.
+# board.STEMMA_I2C() claims and manages I2C_POWER itself, and if
+# another peripheral gets there first the bus comes up with no pull
+# ups, which surfaces as a wiring error rather than a power one.
+def open_air_sensor():
+    """Bring up the STCC4 on the STEMMA QT port."""
+    try:
+        import adafruit_stcc4  # pylint: disable=import-outside-toplevel
```
BlitzCityDIY: please include all imports at the top. as discussed in previous projects, imports for sensors used in the build should not be try/except

MeshFruit_EInk_Message_Board/mesh_sensors.py
```diff
@@ -0,0 +1,185 @@
...
+        return None
+
+
+sensor = open_air_sensor()
+
+
+def open_battery():
+    """Open the onboard MAX17048 fuel gauge."""
+    try:
+        import adafruit_max1704x  # pylint: disable=import-outside-toplevel
```
BlitzCityDIY: same here, all imports at the top

MeshFruit_EInk_Message_Board/mesh_sensors.py
```diff
@@ -0,0 +1,185 @@
...
+
+        gauge = adafruit_max1704x.MAX17048(board.STEMMA_I2C())
+        print("battery gauge up")
+        return gauge
+    except (ImportError, ValueError, RuntimeError, OSError) as gauge_error:
+        print("battery gauge unavailable:", gauge_error)
```
BlitzCityDIY: having these sensors be optional creates a lot of extra code because you're writing code that utilizes them while also having checks in case they weren't initialized. for learn projects i feel like whatever you are documenting in the build or showing as a feature should assume to be used

MeshFruit_EInk_Message_Board/mesh_sensors.py
```diff
@@ -0,0 +1,185 @@
...
+                co2,
+                round(degrees),
+                unit,
+                round(humidity),
+            )
+        except (RuntimeError, OSError) as sensor_error:
+            if attempt == AIR_READ_ATTEMPTS - 1:
+                print("air read failed:", sensor_error)
+                return None
+            # A power blip drops the sensor back to idle, where it
```
BlitzCityDIY: could this be being caused by light sleep?

MeshFruit_EInk_Message_Board/mesh_weather.py
```diff
@@ -0,0 +1,231 @@
...
+import os
+
+WEATHER_INTERVAL = 1800  # seconds between fetches
+
+
+# Open-Meteo is free and needs no API key. Coordinates come from
+# settings.toml so the guide does not hardcode a location.
+WEATHER_URL = (
+    "https://api.open-meteo.com/v1/forecast"
+    "?latitude=%s&longitude=%s"
```
BlitzCityDIY: i'd recommend an f-string here to pass the settings.toml variables

MeshFruit_EInk_Message_Board/mesh_weather.py
```diff
@@ -0,0 +1,231 @@
...
+    86: "Snow showers",
+    95: "Thunderstorm",
+    96: "Thunderstorm",
+    99: "Severe storm",
+}
+
+
+def describe_code(code):
+    """Human readable form of a WMO weather code."""
+    return WMO_CODES.get(code, "Code %s" % code)
```
BlitzCityDIY: f-string here too

MeshFruit_EInk_Message_Board/mesh_weather.py
```diff
@@ -0,0 +1,231 @@
...
+        return None
+
+    try:
+        response = session.get(ZIP_URL % zip_code, timeout=20)
+        data = response.json()
+        response.close()
+        place = data["places"][0]
+        location = (place["latitude"], place["longitude"])
+        print("weather: %s resolved to %s" % (zip_code, place["place name"]))
+        return location
```
BlitzCityDIY: what is the purpose of using Zippopotam? it looks like it just returns latitude, longitude, city and state based on zip code?

MeshFruit_EInk_Message_Board/mesh_weather.py
```diff
@@ -0,0 +1,231 @@
...
+    no reason to keep it running between updates.
+    """
+    ssid = os.getenv("CIRCUITPY_WIFI_SSID")
+    password = os.getenv("CIRCUITPY_WIFI_PASSWORD")
+    if not ssid:
+        print("weather: no wifi credentials in settings.toml")
+        return None
+
+    try:
+        import wifi  # pylint: disable=import-outside-toplevel
```
BlitzCityDIY: please have imports at the top

MeshFruit_EInk_Message_Board/mesh_weather.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def fetch_weather(use_fahrenheit=True):  # pylint: disable=too-many-locals
+    """Pull a short forecast from Open-Meteo. Returns a dict or None.
+
+    WiFi is brought up only for the fetch and shut down afterwards:
+    the ESP32-S3 radio sits close to the LoRa front end, and there is
+    no reason to keep it running between updates.
+    """
+    ssid = os.getenv("CIRCUITPY_WIFI_SSID")
```
BlitzCityDIY: the `getenv` calls should be towards the top of the code so that they are declared variables and not inside of functions.

MeshFruit_EInk_Message_Board/settings.toml.example
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: 2026 Pedro Ruiz for Adafruit Industries
```
BlitzCityDIY: i think we've discussed on past projects to not include a settings.toml file because it can contain secrets but also where it is named settings.toml.example then users will have to rename it but their board will also already have a settings.toml file since the CIRCUITPY drive creates on when CP is installed. settings.toml should just be documented in the guide

BlitzCityDIY (changes requested):
i wonder if fonts, sensors and weather could just be included in code.py? i think that could clean a lot of this up and make it so that you can call things more directly rather than relying on globals inside of functions. other comments are below

MeshFruit_EInk_Message_Board/settings.toml.example
videopixil: @BlitzCityDIY for the modules merging back into code.p - it would directly conflict with the CI already hit. Merging back puts it over the 1000-line pylint limit that forced the split in the first place - should it be a # pylint: disable=too-many-lines ?

BlitzCityDIY: i understand but its more a structure issue and i'm curious if the other edits are taken care of if that would cut down on the length. 

learn guide circuitpython code tends to follow this pattern from top to bottom:
library imports
hardware and module instantiation
variable declaration
methods
the loop

if there are additional python files its usually because of a custom class or handling graphics, things like that. i'd at a minimum like to have the sensors in the main code.py

BlitzCityDIY: this is a great example of code structure for a project like yours: https://github.com/adafruit/Adafruit_Learning_System_Guides/blob/main/MagTag/MagTag_Weather/openmeteo/code.py

MeshFruit_EInk_Message_Board/mesh_icons.py
videopixil: @BlitzCityDIY Went back and forth on this one, but stuck with vectorio.

Those sheets are three-tone — black outline, mid-gray fill, white background. This panel is tri-color, so there's no gray to map the fill to: each cloud would come out either a solid black blob or an empty outline. The gray is doing most of the readability work, especially at 20px.

Drawing them also lets severe weather render in red. The thunderstorm bolt uses the same accent as the newest message and the current conditions row, so color carries meaning rather than decoration.

On cost, vectorio is built into CircuitPython — zero flash. The sheet route needs adafruit_imageload plus both BMPs, and the 70px sheet is ~23KB on its own. This board is tight on a 4MB part; I already had to trim the fonts to printable ASCII to fit.

If you'd rather have consistency with the other weather guides I'll switch — but I think a tri-color redraw of the sheet would be the better fix, and I'd be happy to make one.

MeshFruit_EInk_Message_Board/mesh_sensors.py
videopixil: No. Claiming I2C_POWER manually conflicted with board.STEMMA_I2C(), which already manages that pin. Taking it first made the bus come up with no pull-ups, reported as "No pull up found on SDA or SCL."

MeshFruit_EInk_Message_Board/mesh_sensors.py
videopixil: No, that code is gone now. The real cause was reset=board.D8 and busy_pin=board.D7 in the display setup corrupting I2C. The Feather Friend doesn't use those pins.

MeshFruit_EInk_Message_Board/code.py
videopixil: The three views have entirely different layouts, so persistent labels would mean keeping three sets alive and toggling them. With a full-panel refresh every 180 seconds there's no performance gain

MeshFruit_EInk_Message_Board/mesh_icons.py
BlitzCityDIY: you could redo the color mapping to work with this display like in this guide: https://learn.adafruit.com/preparing-graphics-for-e-ink-displays/overview

videopixil:
okay sweet thanks @BlitzCityDIY went through all of these.

On the power dip comment: no dips, it was claiming I2C_POWER manually when board.STEMMA_I2C() already manages it. Taking it first left the bus with no pull-ups, which reports as a wiring error.

On light sleep causing the sensor failures: no, that was reset=board.D8 / busy_pin=board.D7 in the display setup corrupting I2C. The Feather Friend doesn't use those pins, and that code is gone now.

On rebuilding vs updating .text: I removed the globals, but I'm still rebuilding the group per refresh. The three views have different layouts and the message list varies in count and font size, so persistent labels would mean a fixed pool with .hidden toggling. With a 180s full refresh there's no perf case for it — happy to do it if you'd still prefer.

Fonts are now included in the bundle.

For the is round, staying with the vector icons to keep it moving as Limor requested.

MeshFruit_EInk_Message_Board/mesh_weather.py
```diff
@@ -0,0 +1,178 @@
...
+    "Waxing gibbous",
+    "Full moon",
+    "Waning gibbous",
+    "Last quarter",
+    "Waning crescent",
+)
+
+# Days between new moons.
+
+
```
BlitzCityDIY: duplicate comment

BlitzCityDIY (changes requested):
okay cool, a lot better. just one duplicate comment to remove. going forward i think it would be helpful if you find a project on learn that is similar to what you're working on and share it with claude. that way there is more context for expected style and formatting

BlitzCityDIY (approved):

MeshFruit_EInk_Message_Board/mesh_weather.py
videopixil: yay thanks so much @BlitzCityDIY we did reference existing work throughout — the Star Trek Data Dispenser for the STCC4 (which is what got the sensor working, after single-shot mode failed), the MagTag weather guide for the forecast approach, the MagTag project selector for button-driven views, and the custom fonts guide. Those all shaped decisions.

What was missing was a reference for structure specifically — a Learn guide code.py to model the file layout on. Claude had how to talk to hardware, not how to lay out a guide file

BlitzCityDIY: @videopixil okay but i think you also need to be more of a human in the loop and push back on some of the design decisions based on discussions we've had on these PRs previously. there's a lot of design and structure issues that come up repeatedly and when we're doing a code review it should be more you and me shaping the code and not giving my feedback to claude and posting the response from claude as a rebuttal.

TheKitty:
When I'm using an LLM, I look at the output and when it does something I don't want, I tell it to make a rule not to do something, which is saves. Also Rules to do certain things work as well. We can show it good code in the Adafruit GitHub repos as examples but it takes some human in the loop as Liz notes.

MeshFruit_EInk_Message_Board/mesh_weather.py
videopixil: @BlitzCityDIY agreed, I think some of the memory is lost when switching to the latest models, even though I've set parameters to save to memory.

Unfortunately I don't know how to read code as a CAD designer, this was assigned as just an enclosure design project to begin with.

videopixil:
@TheKitty agreed as noted above, but as mentioned I don't know what im doing - this stated as a case design project for me :)

BlitzCityDIY:
@videopixil well don't hesitate to reach out for code help. previously you would ask me to write the code and i could still do that or build on a base that you get running

videopixil:
@BlitzCityDIY okay awesome you got a deal, I'll absolute take you up on that offer! I've totally felt bad for having to add any more projects to your plate, which I see has ended up adding as much with all the edits, apologies again!

BlitzCityDIY:
@videopixil no all good! we're all learning together. and we always work out scheduling for due dates, so don't worry about that. also writing code for projects is fun for me haha

PR #3288, opened 2026-09-12
Update code.py for Meshfruit class API

videopixil:
- Instantiate Meshfruit once, set mesh.packet per packet and read
  properties rather than calling module functions
- Channel filtering now happens before decryption, since the class
  decrypts lazily on first access to plaintext
- Add a small decode_name helper, as user returns raw bytes
- Tested on hardware: messages decode, NodeInfo short names resolve,
  and examples/meshfruit_simpletest.py receives and prints

### DIFF
--- a/MeshFruit_EInk_Message_Board/code.py
+++ b/MeshFruit_EInk_Message_Board/code.py
@@ -20,7 +20,7 @@ import os
 import time
 
 import adafruit_max1704x
-import adafruit_meshfruit as meshfruit
+import adafruit_meshfruit
 import adafruit_rfm9x
 import adafruit_ssd1683
 import adafruit_stcc4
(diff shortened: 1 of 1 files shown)


TheKitty (approved):

videopixil:
Update for merged Meshtastic_Compatible API

- Instantiate Meshtastic_Compatible to match the merged library naming
- Cycle all three views on a single press; the long press was unreliable
  because the button was only sampled about twice a second
- Sample the button roughly ten times a second so quick taps register
- Redraw every 30 minutes regardless of traffic, so battery, air and
  weather readings do not sit stale with a quiet mesh
- Tested on hardware

BlitzCityDIY (approved):
looks good! is this ready to merge @videopixil ?

videopixil:
@BlitzCityDIY yes tested and ready to merge 🙌

PR #3291, opened 2026-09-21
Handle request retries and rebuild session after failure

videopixil:
- Catch adafruit_requests.OutOfRetries, which inherits from Exception
  and was crashing the board after an ETIMEDOUT
- Drop the cached session on any fetch failure so the next attempt
  builds a fresh socket pool instead of reusing a dead one


videopixil:
@BlitzCityDIY @TheKitty one more edit to fix some timeout failures

BlitzCityDIY (approved):

